PR #11, opened 2015-01-27
[fixed bug] callback now called after write() did it's job.

tillmannheigel:
callback was called right after starting the write()-function. the correct behavior afaik should be to call a callback after a jobs has been done. i had some loading problems on a webpage with a file cache.


oliver-moran:
Looks good. Thanks for that!

PR #15, opened 2015-03-22
Add blit (copy) functionality

gcq:
Also adds a function that returns the index of the pixel in the bitmap given its coordinates

    var lenna = new Jimp("lenna.png", function(){
	    this.blit(this, 247, 235, 98, 136, 0, 0)
	    .write("lennatest.png");
    });

The snippet above copies lenna's face to 0,0 and writes that file.

I've not tested this in much depth, but from what i've seen, corner cases or out of bounds modifications don't break the lib. I'm all for sanity checks though if you want them.

### DIFF
--- a/jimp.js
+++ b/jimp.js
@@ -164,6 +164,16 @@ Jimp.prototype.scan = function (x, y, w, h, cb) {
 };
 
 /**
+ * Returns the pixel index in the bitmap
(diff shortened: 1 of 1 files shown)


oliver-moran:
Thanks, GCQ.

Issue #25, opened 2015-07-14
Allow setting compression and filter for png files


selaux:
First time using your library and I have to say it's great.

Now to the point: As it's possible to set the quality of a jpg file, it would be cool if it's also possible to set some options for pngs, in my case the ```deflateLevel``` and ```filterType``` options if node-png would be required. You think that this would be a good addition?

If yes: What would the functions that implement this look like? I might take a shot at implementing if you'd like.

oliver-moran:
Sounds like a great idea! And I see now that node-png supports it.

If you do a fork and follow a similar pattern used for JPEGs I'll pull in
your changes.

neophob:
I think it make sense to create a separate function to define png specific encoding options - as there are two options. is that ok?

oliver-moran:
Makes sense.

oliver-moran:
This should be pretty simple now that we are using pngjs2: https://www.npmjs.com/package/pngjs2

oliver-moran:
This is all done now. Thanks a lot!

PR #27, opened 2015-08-18
add ordered dithering

neophob:
hi, this function adds ordered dithering to jimp and combined color reduction for RGB565. maybe its too specific, then I would split this function up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Once the callback has fired, the following methods can be called on the image:
     image.brightness( val ); // adjust the brighness by a value -1 to +1
     image.contrast( val ); // adjust the contrast by a value -1 to +1
     image.posterize( n ); // apply a posterization effect with n level
+    image.ordereddithering656(); //ordered dithering the image and reduce color space to RGB565
 
 (Contributions of more methods are welcome!)
 
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Once the callback has fired, the following methods can be called on the image:
     image.brightness( val ); // adjust the brighness by a value -1 to +1
     image.contrast( val ); // adjust the contrast by a value -1 to +1
     image.posterize( n ); // apply a posterization effect with n level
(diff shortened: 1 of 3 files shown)


oliver-moran:
Sounds great. I might just change the method name (it's pretty long).

You've updated the PNG library. I've no objections to that in principle. Just wondering what's better, etc. before I pull it in since it's a fundamental change?

neophob:
Hi, yes feel free to update the name. The new png library is api compatible with the current version - but it's maintained and has more features - see https://github.com/lukeapage/pngjs2#comparison-table 
I'll create a new pull request which will allow to set specific png encoding options soon.

oliver-moran:
> I'll create a new pull request which will allow to set specific png encoding options soon.

That would be great. Have you seen this:
https://github.com/oliver-moran/jimp/issues/25

neophob:
Yes I do - I'll see what I can do - I'll keep this discussion in the other issue.

oliver-moran:
Merged.

oliver-moran:
@neophob I've called the method `image.dither64()`. Is that OK with you?

neophob:
Thanks, well I thinh dither64 is a bit missleading, I would call it dither16 (because the image data is 16bit) or to make it more clear call it dither565

oliver-moran:
Doh! Yes, 16-bit is what I meant! In my mind I was thinking 16-bit was more "user friendly" than 5-6-5. If you think `dither565` is better then, yup, let's go with that. I wanted to check with you before publishing.

I was also thinking it would be nice to have a suit of dithering methods, like here:
https://en.wikipedia.org/wiki/List_of_monochrome_and_RGB_palettes#Regular_RGB_palettes

That was another reason for `dither16` (so we could have `dither3`, `dither6`, etc. in time).

neophob:
I think maybe it would make sense to add more dithering methods yes

oliver-moran:
Changed to `image.dither565()` (and I made an undocumented reference of `image.dither16()`)

PR #28, opened 2015-08-19
remove synchronous file io

mgh:
`getMIMEFromPath` previously did a `ReadChunk.sync`, which throws an
exception if you're out of file descriptors (or some other error
occurs). Changing this to asynchronous with an empty mime type on
failure. It's now pretty easy to handle EMFILE errors with something
like graceful-fs (which catches and delays async io EMFILE errors).

### DIFF
--- a/jimp.js
+++ b/jimp.js
@@ -48,10 +48,12 @@ function getMIMEFromBuffer(buffer) {
     else return "";
 }
 
-function getMIMEFromPath(path) {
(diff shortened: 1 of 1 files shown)


oliver-moran:
Looks good. I'll pull it in and test it later this week. Thanks!

oliver-moran:
Done. Thank you!

PR #30, opened 2015-08-20
Added mask function

disordinary:
I've added a mask function at line 374, it's modelled after the blit function. If a pixel has an rgb balance of 0 then make it transparent.

You create a mask like: http://www.sccs.swarthmore.edu/users/05/zap/labs/E27/lab1/images/circle-mask.png and apply it to the image.

### DIFF
--- a/jimp.js
+++ b/jimp.js
@@ -364,6 +364,44 @@ Jimp.prototype.blit = function (src, x, y, cb) {
 };
 
 /**
+ * Masks a source image on to this image, black areas (R + G + B === 0) are made transparent
+ * @param src the source Jimp instance
+ * @param x the x position to blit the image
+ * @param y the y position to blit the image
+ * @param (optional) cb A callback for when complete
+ * @returns this for chaining of methods
+*/
+Jimp.prototype.mask = function (src, x, y, cb) {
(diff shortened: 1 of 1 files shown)


oliver-moran:
Looks really good. I might make one change: setting the destination alpha channel to the average of the mask RGB. That way if all all are 0 then the alpha is transparent. If all are 255 then the alpha is opaque. And if somewhere in between the alpha is semi-transparent.

Like: http://carijansen.com/2004/07/25/layer-mask-from-grayscale-image/

How does that sound?

disordinary:
Yeah, even better.

oliver-moran:
I've modified it so that it masks based on opacity. Fully transparent pixels on the mask will be fully opaque on the source image. And fully opaque pixels on the mask will be fully transparent on the source image.

oliver-moran:
Integrated and modified to mask based on the average pixel value.

gerges:
@oliver-moran, if you mask based on the average pixel value, then creating a circular mask of an image isn't possible, correct? If I have a image with black where I cant the result to be hidden, then that will be averaged with whatever pixel in the source image?

PR #31, opened 2015-08-27
implemented free choice of angle for rotating images

nitram509:
Hi,

I've implemented a second rotation function that is able to rotate by any given angle.
I left the original one in, just because I think it's optimized for special values of angle (multiples of 90°).
Additionally I've added missing semicolons and put IntelliJ IDEA project files on ignore list.

Any feedback is welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 node_modules
 .project
+.idea
+*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ node_modules
 .project
 .idea
 *.iml
+/test/output/*.png
+/test/output/*.jpg
--- a/jimp.js
+++ b/jimp.js
@@ -177,7 +177,7 @@ function parseBitmap(data, mime, cb) {
                     data: new Buffer(data.data),
                     width: data.width,
                     height: data.height
-                }
+                };
                 return cb.call(that, null, that);
             });
             break;
@@ -212,12 +212,12 @@ Jimp.prototype._quality = 100;
  * @param cb (optional) A callback for when complete
  * @returns the new image
  */
-Jimp.prototype.clone = function(cb){
+Jimp.prototype.clone = function (cb) {
     var clone = new Jimp(this);
 
     if (isNodePattern(cb)) return cb.call(clone, null, clone);
     else return clone;
-}
+};
 
 /**
  * Sets the quality of the image when saving as JPEG format
@@ -355,7 +355,7 @@ Jimp.prototype.blit = function (src, x, y, cb) {
 
     var that = this;
     src.scan(0, 0, src.bitmap.width, src.bitmap.height, function(sx, sy, idx) {
-        var dstIdx = that.getPixelIndex(x+sx, y+sy)
+        var dstIdx = that.getPixelIndex(x + sx, y + sy);
         that.bitmap.data[dstIdx] = this.bitmap.data[idx];
         that.bitmap.data[dstIdx+1] = this.bitmap.data[idx+1];
         that.bitmap.data[dstIdx+2] = this.bitmap.data[idx+2];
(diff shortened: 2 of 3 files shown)


oliver-moran:
Sounds fab. That's a great improvement. Thanks! I'll pull it in an probably to a publish over the weekend.

nitram509:
Thank you for this feedback.
I found two things I want to mention, which may prevent this change to be released in a public release (but can be merged in master for further reviews or work).

1) When doing 90 degree rotation, the width/height changes - but not when one does 89.9 degree. the public rotate() function isn't clear about this behaviour.
We should make the API at this point clear and explizit. To achieve that we can provide two rotation functions, or add a parameter where the user can enable/disable rotation, or we can always rotate, as soon as we round to the next 90 degree.
--> What do you think?  

2) Just 5 seconds before reading your comment, I've found a small bug, which causes an exception. currently i'm investigating this (message: cannot read property height of undefined)
I'm taking care of this.

nitram509:
Problem "cannot read property height of undefined" is fixed now.

nitram509:
About the unexpected rotation behavior,
what do you think about making both method public:
rotate(angle, cb) {...} ==> a rotation with angles 0..360 degree, doesn't change the width nor height of the image
rotate90DegreeSteps(angle, cb) {...} ==> a rotation in 90 degree steps, also reflects in changed width and height of the image.

What do you think?

oliver-moran:
What's wrong with updating the width and height regardless of angle? Obviously this will result in areas of the new image with blank pixels (0x00000000) but that probably desirable in circumstances where someone is rotating to an angle other than multiples of 90 degrees.

I'm thinking for, example, of users where two images might be combined using the `composite` method. In that case, compositing one image, rotated at strange angle with lots of blank pixels, over another will produce a nicely formed composition.

On the other hand, if you don't update the width and height on rotation, won't you end up with a clipped image?

nitram509:
I just want to point out, that Jimp.rotate() now has inconsitent behavior (incl. the patch above).
If one does rotate(89.99) it will not change width/height, but when rotate(90) it will.
As a user of Jimp I wouldn't expect that. My wish is to make the behavior consistent.
That's why I proposed two methods instead of one.

Of course, this means that rotating will result in a clipped image. I don't see a problem in that, because it's the same most graphic tools (e.g. Gimp) work. If one needs another behavior,
one can use other transformation or copy functions beforehand.

nitram509:
@oliver-moran 
Our discussion is already paused for a week now.
May I ask you to continue and see if we can work out a solution we both feel good with.

So, what do you think about my previous proposal?

oliver-moran:
Sorry, @nitram509, not ignoring you. Let me pull in your changes and get back to you in a day or two.

oliver-moran:
OK. I've (finally!) merged this in. I've have it in as "rotate2" for now. Let me mull over whether the discussion above for a bit before making it public. But will do that very shortly!

oliver-moran:
OK. I've made the new method public. There is now just one `image.rotate( deg );` method. If the value is a multiple of 90 then the existing ("simple") function is used. If the value is not a multiple of 90 then the new ("advanced") function is used.

I've altered the advanced function so that its behavior is consistent with the existing function. The clincher was to maintain consistency and backwards compatibility.

However, the method can now be overloaded. So if you call it as `image.rotate( deg, bool );` then you can specify in `bool` if you want the image resized (`true`) or not (`false`).

* * *

There is a seemingly unrelated bug in the test script when trying to manipulate two clones images at the same time. I'll dig into that before releasing.

Thanks for your fantastic contribution!

nitram509:
thanks you.

oliver-moran:
Fix the bug. Should be all good now.

nitram509:
I'm curious - do you plan to release soon?

oliver-moran:
Released now. There were an awful lot of changes in this release. (Too many for one release, to be honest.) I wanted to test it before pushing the button.

PR #34, opened 2015-09-13
Add Option for PNG export: RGB or RGBA

neophob:
this patch adds support for png image type 2 (RGB) and type 6 (RGBA).


oliver-moran:
Thanks! I've pulled this in but changed the method name for usability. It can be accessed through: `image.setAlpha( true | false );`

Issue #38, opened 2015-09-28
Is a command line interface out of scope for this project?


Maximilianos:
I really want to be able to do something like this:
```
npm i jimp -g
```
and then
```
jimp resize --crop image.jpg 420x130 600x240 1024x600
```

Is this something other people want?

oliver-moran:
Definitely not out of scope! My original use case was a build script. A CLI sounds like a great idea.

Maximilianos:
Awesome, will try and get it started this week.

oliver-moran:
Cool. Let me know when you have the smallest publishable piece done. I'll push that and we can start building from there.

Maximilianos:
Ok

riyadhalnur:
:+1: to a CLI version. @Maximilianos would be more than happy to help out if you want

oliver-moran:
I've pulled the CLI into the main branch but will leave it undocumented while underdevelopment.

What say we develop it like that and when we're happy with the API we can begin developing documenting it as "beta"?

Maximilianos:
Sure. will try to work on it a bit more this week, I want to add properly handling resizing/rotating/etc.. a whole dir of images. currently it only handles commands on single images

iamstarkov:
Sindre Sorhus started separate module itself from its cli tool, basically because not all module consumers need cli.

iamstarkov:
I also think its a good idea

Maximilianos:
That makes sense, I also think it is a good idea to separate them

oliver-moran:
So do you think a new project is the way to go? That would make sense IMHO too.

Maximilianos:
Yeah, adding `-cli` is the typical thing done. So a new project `jimp-cli` v. 0.1.0 would probably be the way to go.

oliver-moran:
OK. Do you want me to make a new project here (as a central location) and give you read/write access? Or do you want to start a new project elsewhere?

oliver-moran:
I'll pull the CLI stuff into a separate project over the weekend.

oliver-moran:
Split into a separate project here: https://github.com/oliver-moran/jimp-cli

NPM project is here: https://www.npmjs.com/package/jimp-cli

I've added @Maximilianos as a collaborator here. I presume it's possible to do the same on NPM.

iamstarkov:
@oliver-moran on the right side under the secttion "Collaborators" will "+" icon. Use it
[...]

PR #41, opened 2015-09-30
Cli api

Maximilianos:
Reference: #38

Added necessary scaffold for the Jimp command line api. I am using [commander.js](https://github.com/tj/commander.js) to manage the command line commands and their options.

Added basics for one task to do what was described in the issue ( #38 ):

```
jimp resize test/lenna.jpg 400x200 300 --crop --out-file test/lenna
```

should create two images using the `Jimp.prototype.cover` method, in the `test` directory:
- lenna-400x200.jpg
- lenna-300x300.jpg

Already added additional options for specifying the resize type for each image size, eg:
```
jimp r test/lenna.jpg fl400x200 cr300 cn200x500 -o len
```
will create, in the root directory:
- len-400x200.jpg (using the `resize` method)
- len-300x300.jpg (using the `cover` method)
- len-200x500.jpg (using the `contain` method)

*Developing:* run `npm link` in the Jimp root folder, then you can run `jimp ...` from anywhere in your system to test.

I am still working on this and if merged should be merged into a separate branch until we stabilize the api. Let me know what you think!


cli/index.js
```diff
@@ -0,0 +1,20 @@
...
+	.description('Image manipulation in JavaScript.');
+
+program
+	.command('resize <image> <sizes...>')
+	.alias('r')
+	.description('Resize an image')
+	.option('-c --crop', 'Hard crop an image to the specified sizes')
+	.option('-f --fill', 'Image will be stretched to specified sizes')
+	.option('-o --out-file <name>', 'Name of the output file. Size strings will be appended to this file name')
+	.option('-d --out-dir <dir>', 'Path to the output directory')
```
Maximilianos: Note: I haven't implemented the `--out-dir` flag yet, but the idea is that you would give `jimp` an entire directory of images and it would go through resizing all of them as per the resize instructions and then it would spit them out into the `--out-dir` directory.

oliver-moran:
Cool. I hope to pull this in this evening.

Bit busy IRL for the next few days so if it's not in tonight, it will be next Monday.

cli/index.js
Maximilianos: basic impl of `--out-dir` added see next commit

Maximilianos:
added the rotate command because I needed it at work today. 
abstracted a bit of shared functionality into a utils file

Maximilianos:
added the rotate command because I needed it at work today. 
abstracted a bit of shared functionality into a utils file

Maximilianos:
Hey, any thoughts on this? did you get the time to check it out?

oliver-moran:
I've merged it. Looks great.

For simplicity, I've merged it into the main branch but will leave it undocumented while n development until you're satisfied the API is becoming stable. After that we could document it but flag it as being in development?

PR #43, opened 2015-10-01
Add PNG options filterType and deflateLevel

selaux:
This is a proposal for https://github.com/oliver-moran/jimp/issues/25.

Currently they are two separate methods. Not sure wether one method would be better, something like ```image.pngOptions({ filterType: 1, deflateLevel: 0 })```


oliver-moran:
Thanks, @selaux. I've pull this in. It will be in the next release.

PR #44, opened 2015-10-02
Better greyscale using ITU Rec 709 luminance values

ceautery:
See https://en.wikipedia.org/wiki/Rec._709#Luma_coefficients for reference. Weighting each RGB value equally creates a grey image where blues from the source image are too bright and greens too dim. This change helps pixels in the greyscale image retain the apparent luminance they had in the source image.

### DIFF
--- a/jimp.js
+++ b/jimp.js
@@ -902,13 +902,13 @@ Jimp.prototype.blur = function (r, cb) {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


oliver-moran:
Thanks a lot, @ceautery. I've pulled this in. I'll release it shortly.

PR #48, opened 2015-10-13
return when calling throwError

mgh:
If `throwError` calls a callback function and does not throw itself, execution previously continued. By adding return statements, we prevent further errors from being thrown.

### DIFF
--- a/index.js
+++ b/index.js
@@ -96,7 +96,7 @@ function Jimp() {
 
         if ("undefined" == typeof cb) cb = function () {};
         if ("function" != typeof cb)
-            throwError.call(this, "cb must be a function", cb);
+            return throwError.call(this, "cb must be a function", cb);
 
         this.bitmap = {
             data: new Buffer(w * h * 4),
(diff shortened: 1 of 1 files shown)


oliver-moran:
Sorry now for being thick but I don't understand this.

My intuition says it **looks** good but can you give me an example of before and after? I'll pull it in then.

mgh:
sure! so looking at the diff around line 149:

```diff
             FS.readFile(path, function (err, data) {
-                if (err) throwError.call(that, err, cb);
+                if (err) return throwError.call(that, err, cb);
                 parseBitmap.call(that, data, mime, cb);
```

Consider the case where there was an error reading the file.  Here, `err` will be set and `data` will be `null` (or maybe `undefined`).  Previously, if no `cb` was provided, `throwError` would actually `throw` the error, causing execution to terminate.  However, if you provided a `cb`, `throwError` won't throw.  It will call the `cb` function with `err` as the first argument.  Then it returns, causing `parseBitmap` to get called with invalid data, which then will throw an (uncaught) error itself.  Since the error is already handled, this is undesirable (and also uncatchable except with a global exception handler).  In general, anytime `throwError` gets called, execution of the current function should stop so that invalid data doesn't create future errors.

Does that make sense?

oliver-moran:
Ahh! Sorry for being slow. That's real nice addition/fix. Thanks a lot!

oliver-moran:
This is published on NPM. Thanks for the "catch"!

PR #50, opened 2015-10-13
Color modifications: hue, saturation, lightness and more

rustamli:
Added a way to apply color modifications to images. Most of modifications are done through [TinyColor](https://github.com/bgrins/TinyColor), which is a small open source color manipulation library. However, I have added few of my custom modifiers as well.

Instead of having `image.hue()`, `image.saturate()` etc. as separate methods, I decided to have just one `image.color()` method which handles multiple modifications at once:

```javascript
var output = new Jimp('dice.png', function (err, image) {
  this.resize(600, Jimp.AUTO).color([
    { apply: 'hue', params: [ -90 ] },
    { apply: 'lighten', params: [ 50 ] },
    { apply: 'xor', params: [ '#06D' ] }
  ]).write(filePath.replace('.png', 'dice--out.png'));
});
```

Below is the list of available modifiers:

modifier | description
------------- | -------------
**lighten** {amount} | Lighten the color a given amount, from 0 to 100. Providing 100 will always return white (works through [TinyColor](https://github.com/bgrins/TinyColor))
**brighten** {amount} | Brighten the color a given amount, from 0 to 100 (works through [TinyColor](https://github.com/bgrins/TinyColor))
**darken** {amount} | Darken the color a given amount, from 0 to 100. Providing 100 will always return black (works through [TinyColor](https://github.com/bgrins/TinyColor))
**desaturate** {amount} | Desaturate the color a given amount, from 0 to 100. Providing 100 will is the same as calling greyscale (works through [TinyColor](https://github.com/bgrins/TinyColor))
**saturate** {amount} | Saturate the color a given amount, from 0 to 100 (works through [TinyColor](https://github.com/bgrins/TinyColor))
**greyscale** {amount} |   Completely desaturates a color into greyscale (works through [TinyColor](https://github.com/bgrins/TinyColor))
**spin** {degree} |    Spin the hue a given amount, from -360 to 360. Calling with 0, 360, or -360 will do nothing - since it sets the hue back to what it was before. (works through [TinyColor](https://github.com/bgrins/TinyColor))
**hue** {degree} |   Alias for **spin**
**mix** {color, amount} |   Mixes colors by their RGB component values. Amount is opacity of overlaying color
 **tint** {amount} | Same as applying **mix** with white color 
 **shade** {amount} | Same as applying **mix** with black color 
 **xor** {color} | Treats the two colors as bitfields and applies an XOR operation to the red, green, and blue components
 **red** {amount} | Modify Red component by a given amount 
 **green** {amount} | Modify Green component by a given amount 
 **blue** {amount} | Modify Blue component by a given amount

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ var PNG = require("pngjs2").PNG;
 var JPEG = require("jpeg-js");
 var BMP = require("bmp-js");
 var MIME = require("mime");
+var tinycolor = require("tinycolor2");
 var Resize = require("./resize.js");
 var StreamToBuffer = require('stream-to-buffer');
 var ReadChunk = require('read-chunk'); // npm install read-chunk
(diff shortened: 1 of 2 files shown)


oliver-moran:
Sounds great! Loads of cool stuff there. I've pulled it in. Thanks a lot!

I might add a small number of short-cut methods like `image.darken()` etc. just because we can get them for free now!

Thanks so much! It's awesome.

rustamli:
Thanks for merging @oliver-moran 

Regarding short-cut methods there is something to consider:

Running multiple modifiers with single `image.color()` call would be better for performance than running short-cuts separately (e.g. `image.hue(60).darken(40)`), because changes would happen inside single `image.scan()` in one go. As long as we make it clear on documentation, short-cut methods should be fine. A way to solve this would be to collect all modifiers and apply them together before `image.write` calls, but not sure if it is worth it

oliver-moran:
I agree on all points.

I suggest in the documentation that we document the main `color` method first. Then mention that shortcuts exist to perform these functions one at a time.

rustamli:
sounds good :+1:

husayt:
:+1: this is great indeed. thanks

oliver-moran:
This is published on NPM. Thanks for the great work!

PR #52, opened 2015-10-14
Stop the test directory from being published to NPM

natecavanaugh:
Hi there,
I'm sending this pull request, because when we pull down your module from NPM, the largest chunk of the total payload is from images in the test directory (around 7.5MB of a total of 13.6MB). So ignoring this directory will make the payload less than half the size it currently is.
Would you be okay with merging this in?

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,4 @@
+.project
+.idea
+*.iml
+test/


oliver-moran:
Seems entirely sensible :-) Looking into other optimization so this won't make an NPM until the weekend.

oliver-moran:
Released to NPM just now. Thanks!

natecavanaugh:
Thanks a lot for merging it!

PR #63, opened 2015-11-07
[New Feature] Blit Advanced

webcaetano:
Hi, 

Hi had a problem and find this solution very optimizing for me... 
It changes an [obfuscate image script](https://github.com/webcaetano/image-scramble) to process an 400x400 pixels image from 40 seconds to 120ms.

The source code 

```javascript
Jimp.prototype.blitAdv = function (src, x, y, srcx, srcy, srcw, srch, cb) {
	if ("object" != typeof src || src.constructor != Jimp)
		return throwError.call(this, "The source must be a Jimp image", cb);
	if ("number" != typeof x || "number" != typeof y)
		return throwError.call(this, "x and y must be numbers", cb);

	// round input
	x = Math.round(x);
	y = Math.round(y);

	var that = this;

	src.scan(0, 0, srcw ? srcw : src.bitmap.width, srch ? srch : src.bitmap.height, function(sx, sy, idx) {
		var dstIdx = that.getPixelIndex(x+sx, y+sy);
		var ix = that.getPixelIndex(srcx ? srcx+sx : sx, srcy ? srcy+sy : sy);
		that.bitmap.data[dstIdx] = this.bitmap.data[ix];
		that.bitmap.data[dstIdx+1] = this.bitmap.data[ix+1];
		that.bitmap.data[dstIdx+2] = this.bitmap.data[ix+2];
		that.bitmap.data[dstIdx+3] = this.bitmap.data[ix+3];
	});

	if (isNodePattern(cb)) return cb.call(this, null, this);
	else return this;
};
```

It's basically a blit function with source parameters. 
Before this the closer solution is use jimp.crop and then paste using  jimp.blit
Now you can just use jimp.blitAdv , that works like ctx.drawImage canvas

Hope you guys like it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -709,8 +709,8 @@ Jimp.prototype.blit = function (src, x, y, cb) {
  * @param src the source Jimp instance
  * @param x the x position to blit the image
  * @param y the y position to blit the image
- * @param srcx the x position to blit the image
- * @param srcy the y position to blit the image
+ * @param srcx the x position in the source image
+ * @param srcy the y position in the source image
  * @param srcw the width size of the source image
  * @param srch the height size of the source image
  * @param (optional) cb a callback for when complete
--- a/index.js
+++ b/index.js
@@ -83,7 +83,7 @@ function Jimp() {
         var w = arguments[0];
         var h = arguments[1];
         var cb = arguments[2];
-        
+
         if ("number" == typeof arguments[2]) {
             this._background = arguments[2];
             var cb = arguments[3];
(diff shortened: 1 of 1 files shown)


oliver-moran:
I've added this to the latest version. Rather than adding an "advanced blit", I extended the normal blit to take optional crop parameters. Will release tonight.

webcaetano:
Nice, :+1: :tada: 
Thanks

oliver-moran:
That's available on NPM now. Thanks for the contribution!

PR #77, opened 2015-12-10
Fix up README

haydenbleasel:


oliver-moran:
Thanks, I've pulled this in. It will be in the next release.

PR #79, opened 2015-12-13
Web Worker Support

strandedcity:
A Browserify build of Jimp for the browser. Full documentation of use in the browser is in its own readme at https://github.com/strandedcity/jimp. Working demo available at http://strandedcity.github.io/jimp/browser/.

I've attempted to keep the diffs on existing files (readme.md and index.js) as small as possible. Almost all of the changes are in a new "/browser" directory, including a build script that allows new versions of jimp.js and jimp.min.js to be created any time.

All existing tests pass without changes when I ran them just now.

I considered more invasive approaches to the "this code works in a browser, this code doesn't" problems, such as dividing index.js into sub-modules for "node" and "browserify". I decided in the end that this annotation approach was not too messy, and keeps the existing files in tact.

Also included in this PR: a new version of the .contain() function that does the same thing, but outputs a version of the image without the background area.

Interested to hear your thoughts, and thanks again for the excellent library! It's exciting to be able to do this so performantly in the browser.


oliver-moran:
Cool.

Give me a few days to get my head around this then I'll pull it and make a release.

strandedcity:
Sounds good. Happy to make any changes you think might improve usefulness or clarity!

The build process is a little bootstrapped.... it may be be better to run it in a "postinstall" script so that npm would do the build, and browserify wouldn't have to be installed globally. But since the browserified code is run in a completely separate environment and has nothing to do with installing Jimp for use in node, I figured it was just as good separate. Even better maybe, since I anticipate people will probably make their own builds with just the functionality they need to reduce script size. 90kb includes everything in Jimp, but that's big for a gzipped JS file. I saved substantially by leaving only the resize, crop, and encode/decode functionality in place.

I am, of course, using this code in another project I'm working on. If I see any big gaps or improvement opportunities myself as work continues, I'll certainly let you know.

strandedcity:
Just a small heads-up here, the IE compatibility (or lack thereof) kept bothering me. Looks like one of Jimp's dependencies (Quantization) uses ES6 syntax (such as "const"). That breaks IE10 compatibility, but I think it's relatively easy to address. I may need to add a commit to this branch that adds Babel (or babelify) to the toolchain. The code included here likely won't change much, however.

strandedcity:
Alright! Sorry for the delay finishing up here. Made several improvements here to get things running smoothly in IE, as well as general error handling improvements. The script's dependencies are now included in package.json, and the bash script is run with a "prepublish" hook, so it should be impossible to get out of sync with the version in the repo. It's also now easy as pie to run the script.

```
$ npm install

> jimp@0.2.20 prepublish /home/phil/PROJECTS/jimp
> ./browser/browserify-build.sh

Browserifying browser/jimp.js...
Translating for ES5...
[BABEL] Note: The code generator has deoptimised the styling of "tmp1.js" as it exceeds the max of "100KB".
Removing Strict Mode.
Adding Web Worker wrapper functions...
Minifying browser/jimp.min.js...
Cleaning up....
```

strandedcity:
Fixes a browserify dependency. Was using the browserify.transform field specified in browserify's docs, but it wasn't working for clean installs. https://github.com/substack/node-browserify

With browserify dependencies specified as separate npm packages, Jimp installs smoothly from clean slate:

```
$ npm cache clean && rm -rf ./node_modules && npm install
```

strandedcity:
hey there, happy New Year! Any questions I can answer to help make all this more familiar? Jimp compiled this way has been working great for me the last couple weeks; I think it's a pretty stable build process at this point. If you prefer keeping Jimp server-side-only I can maintain the fork separately (it has very few changes to existing files) but I think it makes more sense as part of the original project. Does the build process and demo work smoothly for you now?

oliver-moran:
@strandedcity sorry for being so slow getting back to you. I really appreciate this but I'm temporarily mega busy IRL. I cannot promise you that I'll get a moment before march to check this and release it. But I promise I will release it in March at the latest!

strandedcity:
No worries, totally understood. I'll be here when things calm down!

addyosmani:
@strandedcity Excellent work trying to get worker support in here. Is https://github.com/strandedcity/jimp/tree/gh-pages/browser up to date with the latest commits in your fork? When testing in Chrome (desktop and Android) stable or canary, any file uploads result in uncaught buffer errors.

strandedcity:
I'll take a look today. This has definitely been working (I've been testing with PNGs), but I also see what you're seeing now.

strandedcity:
Very strange.... some of the PNGs in test/samples seem to process fine, others result in this buffer error. Seems similar to [this issue](https://github.com/devongovett/browserify-zlib/issues/5), but the fix suggested there doesn't get the job done. This appears to be more than a simple mistake on my part... I'll see if I can pin it down to a bug in the browserify buffer module, or pngjs. The buffer is created correctly (it has the right size and mime type anyways), but this line in index.js triggers the error you're seeing:

```
png.parse(data, function(err, data)
```

Stay tuned.

strandedcity:
Ok, good news and bad news here.

Good news: I think I found a fix! [Give it a whirl](http://strandedcity.github.io/jimp/browser/), and let me know if that works for you? Thanks for the catch!

Bad news: the problem lies deep in the bowels of the Pako Zlib Inflate Library, apparently in one of the [error-checking conditions](https://github.com/nodeca/pako/blob/master/lib/zlib/inflate.js#L1449) as the PNG is decompressed. This is inside jimp > browserify > browserify-zlib > pako > lib > zlib > inflate.js.

I'm not sharp enough on buffers, compression, and PNGs to be able to properly fix that condition. But I do know that if I comment out that line, PNGs work in Jimp for Web Workers (I've just tested a bunch from the [PNG test suite](http://www.schaik.com/pngsuite/)).

To get things working I've deleted that line as browser/lib/jimp.js is being built. I would love to hear of a better solution, but this appears to work. Meanwhile this change only affects the web-worker version of the library since none of the Browserify stuff affects Jimp for node in the least. That's also why the bug only existed in my code, not when running pngjs in node. The pngjs code is right, but its dependencies are traded under the hood by Browserify.

strandedcity:
@addyosmani wanted to make sure you saw there was a solution here.

addyosmani:
@strandedcity Thanks for the heads up! Appreciate you digging down the dependency chain to locate that gnarly bug. I'll give the workaround a spin. Cheers for sharing it :)

index.js
```diff
@@ -1448,6 +1449,27 @@ Jimp.prototype.contain = function (w, h, cb) {
 };
 
 /**
+ * Scale the image to the largest size so that its width and height fits inside the given width and height. Crop to image area.
```
oliver-moran: Please remove this method.

oliver-moran:
OK. my real life is now back to normal and I'm good to pull this in. Before I do, could you remove the parts to do with the `Jimp.prototype.containCropped` method (in `index.js` and `README.md`)? I'll pull it in then and test it.

strandedcity:
Done. Question, though:

Is the problem the functionality, the naming, or its presence in this pull request? It might also be called "aspectFit()" or something like that, but basically I was trying to correct for ```.contain()``` doing two things:

- Scale the image to the largest size so that its width and height fits inside the given width and height.
- Blit that image into the center of a square of the larger of the two dimensions

I found it confusing that ```.contain(300,200)``` outputs a 300px square image.

oliver-moran:
> I found it confusing that .contain(300,200) outputs a 300px square image.

Ah. That's a bug. It should output a 300 x 200 image that contains the original image scaled to fit, with aspect maintained, and centred.

strandedcity:
Oh! Perfect. I'll make a separate pull request to fix that, then. The contents of my deleted function does what you describe.

strandedcity:
Oh, actually not quite. My intent was to scale the image to fit, then return it. If input aspect ratio does not match provided dimensions, the retuned image would be smaller than provided dimensions. My function does not do the letterboxing that contain() does. In any case, it's a separate issue.

PR #82, opened 2015-12-17
Added an optional 'tolerance' parameter to autocrop method (and corrected minor bugs)

marcolino:


oliver-moran:
Looks great. Thanks so much.

I'm mega busy IRL tho until March so I won't be doing a release before then. I will pull it in tho and add it to that release.

Thanks again.

index.js
```diff
@@ -764,22 +800,42 @@ Jimp.prototype.autocrop = function (cb) {
...
+            if (difference > tolerance) {
+                // this pixel is too distant from the first one: abort this side scan
                 break east;
             }
         }
         // this column contains all pixels with the same color: increment this side pixels to crop
         eastPixelsToCrop++;
     }
 
+    colorTarget = this.getPixelColor(w - 1, h - 1); // bottom right pixel color is the target color
```
Iwasawafag: Here's an error. Bottom right pixel is being read, but the value is never used to update `rgba1` variable, and every pixel still gets compared against the top left one.

PR #95, opened 2016-02-17
Replace resize.js with resize method of imagejs for additional interpolation modes

skoa-h:
With the replacement there are now multiple interpolation modes supported.

Default mode is Jimp.RESIZE_BICUBIC.

Modes currently supported by imagejs:
```javascript
Jimp.RESIZE_NEAREST_NEIGHBOR
Jimp.RESIZE_BILINEAR
Jimp.RESIZE_BICUBIC
Jimp.RESIZE_HERMITE
Jimp.RESIZE_BEZIER
```
Adds dependency to imagejs


oliver-moran:
I've done this - but done it a bit differently.

You can now pass an optional resize mode to the resize method, like this:

```js
this.resize(256, Jimp.AUTO, Jimp.RESIZE_BICUBIC) 
```

This is available now from the main branch and I'll release it over the next week or so.

PR #98, opened 2016-03-01
`.contain` does more than expected

Mithgol:
`README.md` [currently](https://github.com/oliver-moran/jimp/blob/2d3da35d0f4791b1c5a3fb12d0dc58a7e1ce456e/README.md#basic-usage) says the following:

```js
image.contain( w, h ); // scale the image to the largest size so that fits inside the given width and height
```

However, in reality not only the `.scale` is performed, but also [background-filled rectangles are added around that image](https://github.com/oliver-moran/jimp/blob/2d3da35d0f4791b1c5a3fb12d0dc58a7e1ce456e/index.js#L1466-L1470) so that the image starts to cover the entire w×h area.

This pull request tones down the behaviour of `.contain`, makes it more README-corresponding.

The old abilities of `.contain` are, however, stored in the new method `.container` and explained anew in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ image.mask( src, x, y );       // masks the image with another Jimp image at x, 
 image.dither565();             // ordered dithering of the image and reduce color space to 16-bits (RGB565)
 image.cover( w, h );           // scale the image so that it fills the given width and height
 image.contain( w, h );         // scale the image to the largest size so that fits inside the given width and height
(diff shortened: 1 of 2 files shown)


oliver-moran:
This is expected behavior. But I'll update the docs to make it clear. It method resizes the image to the specified w × h while mimicing the CSS `background: contain` style. This results in letter boxing (or pillar boxing) if the image aspect ratio changes.

If you wanted trim out the blank spaces afterwards, you could use:

```js
image.contain(500, 500).autocrop().write("image-scaled.png");
```

PR #106, opened 2016-03-11
+1 method: `scaleToFit`

Mithgol:
This method scales the image to the largest size that still fits inside the rectangle that has the given width and height.

It corresponds with the ideas of @strandedcity from #79 and my ideas from #98.


strandedcity:
:+1: 

```scaleToFit``` + ```blit``` provide all the functionality of ```contain``` while keeping the functions more atomic.

oliver-moran:
Thanks. This will be included in the next release, which will happen shortly.

PR #109, opened 2016-03-20
fix npm test

neophob:
fix npm test


oliver-moran:
Merged. Thank you. This will be in the next release.

PR #113, opened 2016-03-29
bug fix: Changing Promise->Promise_ 

wahengchang:
**_TypeError: Promise.defer is not a function
_**

This commit to solve the issue I got recently in jimp, which is Promise overwrites another Promise was using in other library. 

And thx for my bro solving this issue together:
https://github.com/sigshen

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,7 @@ var URLRegEx = require("url-regex");
 if (process.env.ENVIRONMENT !== 'BROWSER') var Request = require('request').defaults({ encoding: null });
 
 // polyfill Promise for Node < 0.12
-var Promise = Promise || require('es6-promise').Promise;
+var Promise_ = Promise || require('es6-promise').Promise;
 
 // logging methods
 
@@ -196,7 +196,7 @@ function Jimp() {
  * @retuns a promise
  */
 Jimp.read = function(src, cb) {
-    var promise = new Promise(
+    var promise = new Promise_(
         function(resolve, reject) {
             cb = cb || function(err, image) {
                 if (err) reject(err);


oliver-moran:
No a fan of this, primarily because I don't understand why it the Promise variable "escaped" from it's closure and overwrote the other promise. Any idea?

s4tori:
I think there are 2 issues.

Firstly, the line `var Promise = Promise || require('es6-promise').Promise;` does not work as expected with node.js.
Because of the `var` statement, the Promise variable is initialized to `undefined`. 
And Jimp always includes `es6-promise`.
```js
console.log(Promise); // undefined
var Promise = Promise || require('es6-promise').Promise; // undefined || require(...)

console.log(Promise); // always 'es6-promise'
```

There is also a "bug" with `es6-promise`, look at this line of code : [es6-promise/polyfill.js#L21-L25](https://github.com/stefanpenner/es6-promise/blob/5df472ec56d5b9fbc76589383852008d46055c61/lib/es6-promise/polyfill.js#L21-L25)
`es6-promise` automatically polyfill some Promise implementations ([bluebird](https://github.com/petkaantonov/bluebird) for example).
See here for more details : [es6-promise#163](https://github.com/stefanpenner/es6-promise/issues/163)

A proper way to get the right Promise implementation is to use the [global](https://nodejs.org/api/globals.html#globals_global) keyword :

```js
var Promise = global.Promise || require('es6-promise').Promise;
```

I made a new Pull Request for this purpose : #128.

oliver-moran:
Closing because the latest pull request resolves the problem. Thank you!

wahengchang:
Well don e!

PR #115, opened 2016-04-01
Add normalization function

arminrosu:
If you accept this incredible, one-time-only offer, jimp could have [normalization](https://en.wikipedia.org/wiki/Normalization_(image_processing)) included today! ;)

You can see it's unbelievable effects below. The bushes maintain their highlights.

sample

The original Lenna is on the left. Notice the yellow overcast.
lenna

As an added bonus, I've optimized the images as well - except the exif samples, of course. I managed to shave 30MB of the repo size.

If someone is so inclined, they could use the `histogram()` function to create an [equalization](https://en.wikipedia.org/wiki/Histogram_equalization) method.


oliver-moran:
An offer I cannot refuse. Thanks. Will be included in the next release.

I didn't pull in the images. A branch conflict meant I couldn't simply merge and a `.npm-ignore` file now leaves the test files out of the release anyway.

arminrosu:
A fine choice, a fine choice indeed.

Issue #75, opened 2015-12-09
Browserified version


strandedcity:
Hi There!

Jimp is terrific. Thanks so much for all the work you've put into it! I've prototyped an addition to the project I thought might be of interest. Some background:

A project I'm working on requires a series of thumbnails of different sizes for each image uploaded by a user. Doing this server-side (which is the typical solution used by most) requires substantial investment in server infrastructure to handle the load with reasonable response times, but doing the same work using `<canvas>` blocks the UI thread in a way that interrupts user's work. This is a very common use case that generates a lot of load around the internet.

Using Jimp + Browserify, I've successfully prototyped a client-side image thumbnailer that runs as a web worker. It turns out that performance is terrific, and the workload of thumbnailing can be successfully distributed to clients. The minified + gzipped file is 300kb, reasonable enough for an application that is by definition graphics-heavy anyways.
- My fork with the quick-and-dirty prototype: https://github.com/strandedcity/jimp
- Demonstration of the web-worker in action: http://strandedcity.github.io/jimp/browser/

(just select any jpg or png image and wait a few seconds, a thumbnail will appear)

Currently, I've hacked together a rather inflexible prototype that involves deleting code to remove dependencies... not good pull request material.

My question is this: if I were to rework my changes so that they are non-destructive to the original project, would you be interested in seeing a pull request for a web-worker version of jimp to be included in the main repo? Seems like something many others would be interested in.

Thanks again for the awesome project!

oliver-moran:
Sounds smashing.

I couldn't get your demo page to work (`Uncaught TypeError: Cannot read property 'width' of undefined`) but I get the idea. And it sounds great.

If you code it, I will pull.

strandedcity:
Cool, will do! I'll watch browser compatibility as well :+1:

strandedcity:
Here we go: https://github.com/oliver-moran/jimp/pull/79

oliver-moran:
Thanks, I've replied on the pull request thread.

oliver-moran:
OK. I pulled it and broke it.

I made one change: https://github.com/oliver-moran/jimp/commit/d2e9582e71e98b395ed308a088f880922f3f0b42#diff-168726dbe96b3ce427e7fedce31bb0bc

I don't think that should have broken it but when I did a `npm install` afterwards the browser example stopped working. Any ideas?

strandedcity:
I saw this at one point also. I think it's because uglifyjs will remove this:

```
[...]

PR #116, opened 2016-04-05
Use Buffer.isBuffer() instead of constructor check

sqrrrl:
Ran in to an issue where using jimp/browser was failing in an odd way and returning false for the buffer checks even though the actual value was a buffer. I don't know if it was the particular environment (babel + webpack) but changing it to use Buffer.isBuffer() instead of comparing the constructor function seemed to do the trick for me.

### DIFF
--- a/index.js
+++ b/index.js
@@ -147,7 +147,7 @@ function Jimp() {
         var that = this;
         Request(url, function (err, response, data) {
             if (err) return throwError.call(that, err, cb);
-            if ("object" == typeof data && Buffer == data.constructor) {
+            if ("object" == typeof data && Buffer.isBuffer(data)) {
                 var mime = getMIMEFromBuffer(data);
                 if ("string" != typeof mime)
                     return throwError.call(that, "Could not find MIME for Buffer <" + url + "> (HTTP: " + response.statusCode + ")", cb);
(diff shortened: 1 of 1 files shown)


oliver-moran:
Merged. Thank you. This will be included in the next release.

strandedcity:
@sqrrrl did you have any issues getting the browser build to run (aside from this check)? We're still puzzling through getting it to run smoothly for @oliver-moran. https://github.com/oliver-moran/jimp/issues/75

sqrrrl:
@strandedcity I ran into a bunch of issues, mostly related to trying to get it to play nice with webpack & babel. A few things I had to do:

- Monkey patch the file as mentioned in this patch
- Use the import loader (in webpack.conf.js I set up a loader rule using: 'imports?require=>false') since it didn't play nicely with webpack out of the box.
- I had to use the unminified version. For some reason, the minified version is still causing errors for me and I haven't dug too far into it. My use was just for a silly toy project, so I wasn't all that motivated to get  fully optimized build going.

That said, still running in to some issues with PNG files not parsing correctly. I don't know if that's specific to the browser version or just if jimp just isn't that robust. Again, not something I spent a whole lot of time on trying to figure out.

strandedcity:
Is this a different build process than the one run in ```browser/browserify-build.sh``` ? I'm not familiar with ```webpack.conf.js```, but I believe it would be another path to running Jimp in the browser, besides using browserify. 

Do you have an example PNG that's causing problems when you try to upload it on the demo page, http://strandedcity.github.io/jimp/browser/ ? Would be helpful to have that as a test case. I'm aiming to use Jimp in a browser context in production code, so I'm invested in getting it 'close enough'.

Thanks!

PR #122, opened 2016-04-15
Catch expressions of jpeg-js

netdown:
Currently expressions thrown by jpeg-js are not catched and they simply break the script.
If a file is corrupted, jpeg-js will throw a message which could be forwarded to the callback function.


oliver-moran:
Thank you. Will include in a release very shortly.

PR #125, opened 2016-04-21
Upgrade jpeg-js

strandedcity:
Upgrade jpeg-js to include the change that fixes color reversal when decoding CMYK jpegs.

https://github.com/eugeneware/jpeg-js/pull/16


oliver-moran:
Thank you. Will include in a new release very shortly.

PR #127, opened 2016-04-22
Slighly less ugly build paths

strandedcity:
Here's one modest improvement to the elegance.... npm is able to pick out the local copy of uglify, but apparently only if it's run through a script defined in package.json. Still not amazing.... thoughts?

https://lostechies.com/derickbailey/2012/04/24/executing-a-project-specific-nodenpm-package-a-la-bundle-exec/


oliver-moran:
Beautiful.

Issue #129, opened 2016-04-26
Error installing jimp, 'node-zlib-backport error'


mirkomilova10:
So, for some reason, out of the blue my nodejs refuses to work as it should. when i run `npm insall jimp` i get a node-gyp error on the module node-zlib-backport used by the dependency pngjs. it returns an error when compiling the source giving a `Users/mirko/.node-gyp/5.11.0/include/node/node_internals.h:6:10: fatal error: 'util-inl.h' file not found
#include "util-inl.h"` error.

Heres the (almost) full terminal log. i know this might not be jimp's problem but i wonder if anyone experienced it before:

```
node-zlib-backport@0.11.15 install /Users/mirko/seriesrocketNotifierApp/node_modules/node-zlib-backport
> node-gyp rebuild

  CXX(target) Release/obj.target/zlib/src/node_zlib.o
In file included from ../src/node_zlib.cc:32:
/Users/mirko/.node-gyp/5.11.0/include/node/node_internals.h:6:10: fatal error: 'util-inl.h' file not found
#include "util-inl.h"
         ^
1 error generated.
make: *** [Release/obj.target/zlib/src/node_zlib.o] Error 1
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:276:23)
gyp ERR! stack     at emitTwo (events.js:100:13)
gyp ERR! stack     at ChildProcess.emit (events.js:185:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:204:12)
gyp ERR! System Darwin 15.4.0
gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /Users/mirko/seriesrocketNotifierApp/node_modules/node-zlib-backport
gyp ERR! node -v v5.11.0
gyp ERR! node-gyp -v v3.3.1
gyp ERR! not ok 
npm WARN install:node-zlib-backport@0.11.15 node-zlib-backport@0.11.15 install: `node-gyp rebuild`
npm WARN install:node-zlib-backport@0.11.15 Exit status 1
```

Running on OS X with node 5.11.0 and npm 3.8.6

oliver-moran:
[...]

PR #130, opened 2016-04-27
printWrapped

Igor1201:
Adds a `printWrapped` function, which measures text width to be printed, continuing to print it on a newline, if necessary. 

~~Also,~~
- ~~fixed [index.js#L1169](https://github.com/oliver-moran/jimp/blob/master/index.js#L1169) threading arrow~~
- ~~updated `pngjs` to `2.3.1`, fixing #129~~

### DIFF
--- a/index.js
+++ b/index.js
@@ -2168,6 +2168,60 @@ function drawCharacter(image, font, x, y, char) {
     return image;
 };
 
+
+/**
+ * Draws a text on a image on a given boundary
+ * @param font a bitmap font loaded from `Jimp.loadFont` command
+ * @param x the x position to start drawing the text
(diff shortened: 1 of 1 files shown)


oliver-moran:
Looks great. Thanks a lot. And sorry for delay getting back to you!

We've got conflicts and I suspect you'd be quicker at fixing them that I would. If you can fix them, I'll pull in and make a release later this week.

Igor1201:
@oliver-moran no problems, man!
I've just solved the conflicts

oliver-moran:
Love it! Thank you.

Before a release, I might integrate the two print method in some way so that a user doesn't have to make a hard decision between the two. But expect it to be out in the wild ASAP.

PR #131, opened 2016-04-27
Define window

strandedcity:
Fixes a couple new issues with the browserify build:
- Choking on ES6 syntax in index.js (not sure why?)
- window is not defined error is thrown when running as a web worker (where ```self``` is the global variable)

See https://github.com/oliver-moran/jimp/issues/75

### DIFF
--- a/browser/browserify-build.sh
+++ b/browser/browserify-build.sh
@@ -37,8 +37,8 @@ cat ../LICENSE;
 echo "*/";
 echo ""; } > tmp.web_license.txt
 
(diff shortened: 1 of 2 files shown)


oliver-moran:
`Node's Buffer() not available in jimp-worker.js`

:-(

strandedcity:
I had this in my diff for this PR, and almost left it in in case the pngjs nonsense continued, but took it out in the interest of being conservative.

I think this gets the job done though:

https://github.com/oliver-moran/jimp/pull/133

PR #133, opened 2016-04-28
Ignore missing dependencies in browserify build

strandedcity:
This was the problem with pngjs when node-zlib-backport was unavailable.


oliver-moran:
Thanks! Merged and will be in the next release.

PR #136, opened 2016-04-29
add xhr method for electron

super-ienien:
Add useXHR static method when executing in atom-electron renderer process.

when we do Jimp.useXHR(true) It replace Request module by the XMLHttpRequest function of the browser to make http request for reading an image via url.

This can be usefull when the electron application is running on a network connected to the internet behind a proxy server. Because the browser is aware of the system proxy configuration and node isn't.

If you think this is an interesting feature, i can add a note in the readme before you merge.Add useXHR static method when executing in atom-electron renderer process.

when we do Jimp.useXHR(true) It replace Request module by the XMLHttpRequest function of the browser to make http request for reading an image via url.

This can be usefull when the electron application is running on a network connected to the internet behind a proxy server. Because the browser is aware of the system proxy configuration and node isn't.

If you think this is an interesting feature, i can add a note in the readme before you merge.

Sorry for the previous PRs my Git fork was messed up

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,38 @@ var URLRegEx = require("url-regex");
 var BMFont = require("load-bmfont");
 var Path = require("path");
 
-if (process.env.ENVIRONMENT !== 'BROWSER') var Request = require('request').defaults({ encoding: null });
+if (process.env.ENVIRONMENT !== 'BROWSER') {
+    var Request = require('request').defaults({ encoding: null });
+    //If we run into electron renderer process, add a method to substitute the request module by xhr method
+    if (process.versions.hasOwnProperty('electron') && process.type === 'renderer') {
+        var RequestXHR = function (url,cb) {
+            var xhr = new XMLHttpRequest();
+            xhr.open( "GET", url, true );
+            xhr.responseType = "arraybuffer";
+            xhr.onload = function() {
+                if (xhr.status < 400) {
+                    try {
+                        var data = Buffer.from(this.response);
+                    } catch (e) {
+                        return cb("Response is not a buffer for url "+url)
+                    }
+                    cb(null, xhr, data);
+                }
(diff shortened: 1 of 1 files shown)


oliver-moran:
Sounds good. Maybe need a refactor of some kind. But in principle looks good.

Is there any time when it would be more useful to use the node Request module when running in an electron application? If not, why not make this the default behavior?

super-ienien:
There is two kind of places where you can run libraries into Electron apps 

The main process wich is just like a node app so you don't have the XMLHttpRequest function (Request should be used here)
And the renderer processes wich is like a node child process but with all the capabilities of a chromium window (here you can use Request or XMLHttpRequest)

I think it could be the default behavior to use XMLHttpRequest when you run Jimp into an Electron's renderer process. I didn't do it because i didn't want to change the default behavior of Jimp, i think that kind of decision belong to its creator.

oliver-moran:
Oh yeah. I only mean if we hit the renderer :-)

You could do a double (triple?) check for the existence of `XMLHttpRequest`, just to be sure e.g.:

    if (process.versions.hasOwnProperty('electron')
        && process.type === 'renderer'
        && typeof XMLHttpRequest === "function") {
            // do stuff here
    }

zeke:
Don't know much about this project or the intent here, but you can also use `window.fetch` which is built into chromium.

https://developers.google.com/web/updates/2015/03/introduction-to-fetch?hl=en

dpbeaumont:
> Don't know much about this project or the intent here, but you can also use `window.fetch` which is built into chromium.
> 
> https://developers.google.com/web/updates/2015/03/introduction-to-fetch?hl=en

This would be good, as when I use Jimp in a cloudflare worker that only has access to the fetch API. I get the following. 

> ReferenceError: XMLHttpRequest is not defined

PR #138, opened 2016-05-07
typofix

mLuby:
iamge -> image


oliver-moran:
Thanks! Will be in the next release!

PR #140, opened 2016-05-16
fix a spell error

xinnix:
fix a spell error


oliver-moran:
Thank you! Merged and will be in the next release.

Issue #142, opened 2016-05-24
Misleading error when `file-type` module returns no MIME type


rifflock:
Hi!
I'm pulling an images from URLs and I've noticed that I occasionally get the `[Error: Unsupported MIME type: ]` result that some other users have posted about.  The image is a JPEG, and it appears that the buffer of the image isn't being properly resolved to its MIME type by the `filte-type` module. 

JIMP returns either a valid MIME or an empty string from the `getMIMEFromBuffer` method, but the two locations I found that call that method do only validation that the response is of type `string`. Would it not make more sense to return a non-string type in the event of failure to find MIME?
```
function getMIMEFromBuffer(buffer) {
    if (FileType(buffer)) return FileType(buffer).mime;
    else return null;
}
```
So that the calling functions will throw the error immediately instead of passing the empty MIMEtype on?

I was also thinking of extending the `getMIMEFromBuffer` to attempt to also get the MIME from the original filename as referenced in #114 

Does that sound like a useful PR to put together?

oliver-moran:
All makes good sense to me. Thank you!

PR #143, opened 2016-05-24
also check file extension

rifflock:
This is my initial fix for #142 
Comments and suggestions welcome.


oliver-moran:
Thanks! I think this might be missing a change further up the file to pass `path` to the updated function but I will add this. Will be in the next release!

PR #145, opened 2016-06-06
fix dithering, improve output quality of the dithered image

neophob:
Hi

I used the dithering method in a real life application with a 16bpp display. the original dithering table works but gives visual artefacts. the new table improve the output quality.


oliver-moran:
Thanks! Merged. I've no way to test this so I'll trust it's OK. Will be in the next release.

PR #155, opened 2016-07-27
add xhr method for electron by default

super-ienien:
Like #136 but XHR is used by default in electron

Now you can choose wich one you wan't


oliver-moran:
Love it. Thank you! Will release over the weekend (probably!).

PR #159, opened 2016-08-16
Ignore color information of completely transparent pixels when resizing images

Iwasawafag:
Fixes artifact reported in haydenbleasel/favicons#123

Black faint on the edges of opaque areas appears because color information of completely transparent pixels is taken into account during calculation of average color, but these pixels could be of any color. Completely transparent black completely transparent red looks identicall, so it's somewhat a common practice to replace RGB values with 0 or 255 for series of completely transparent pixels. Many encoders do this, (probably) to allow for a better compression. 

Two images below illustrate what transparent areas sometimes look like if we make every pixel opaque.

This PR changes both submodules responsible for image resizing in the way that only color data of visible pixels used in average color calculation. Also ~~fixed incorrect channels value in `Resize.prototype.resizeWidthInterpolatedRGBA` method (Iwasawafag/jimp@65158a9) and~~ reduced amount of dupplicated code in `/resize.js`

And here's a comparison table.  6 images, every transparent pixel of which was painted in black, were scaled down from ~1000x1000 to 200x200 px using every resampling method supported. Then results from before and after patch were glued together for every image. The difference is pretty noticeable

|          | Resized to 200x200. Before/After (click to enlarge) |
|----------|-----------------------------------------------------|
| default  |  ~  ~   ~  ~  |
| bezier   |  ~  ~   ~  ~  |
| bicubic  |  ~  ~   ~  ~  |
| bilinear |  ~  ~   ~  ~  |
| hermite  |  ~  ~   ~  ~  |
| nearest  |  ~  ~   ~  ~  |

Here's list of images from the table above in their original size:

- [Check mark](http://i.imgur.com/l93ZrVT.png)
- [CP logo](http://i.imgur.com/OaEaqA4.png)
- [Green cross thing](http://i.imgur.com/Xp6pIIu.png)
- [Cross thing with gradient fill](http://i.imgur.com/0TwxXs1.png)
- [White cross thingy](http://i.imgur.com/y1T1DTw.png)
- [Red cross thingy](http://i.imgur.com/Ss68BIk.png)

### DIFF
--- a/resize.js
+++ b/resize.js
@@ -57,7 +57,55 @@ Resize.prototype.configurePasses = function () {
     }
 }
 
-Resize.prototype.resizeWidthRGB = function (buffer) {
+Resize.prototype._resizeWidthInterpolatedRGBChannels = function (buffer, fourthChannel) {
+    var channelsNum = fourthChannel ? 4 : 3;
+    var ratioWeight = this.ratioWeightWidthPass;
+    var weight = 0;
+    var finalOffset = 0;
+    var pixelOffset = 0;
+    var firstWeight = 0;
+    var secondWeight = 0;
+    var outputBuffer = this.widthBuffer;
+    //Handle for only one interpolation input being valid for start calculation:
+    for (var targetPosition = 0; weight < 1 / channelsNum; targetPosition += channelsNum, weight += ratioWeight) {
+        for (finalOffset = targetPosition, pixelOffset = 0; finalOffset < this.widthPassResultSize; pixelOffset += this.originalWidthMultipliedByChannels, finalOffset += this.targetWidthMultipliedByChannels) {
+            outputBuffer[finalOffset] = buffer[pixelOffset];
+            outputBuffer[finalOffset + 1] = buffer[pixelOffset + 1];
+            outputBuffer[finalOffset + 2] = buffer[pixelOffset + 2];
+            if (!fourthChannel) continue;
+            outputBuffer[finalOffset + 3] = buffer[pixelOffset + 3];
+        }
+    }
+    //Adjust for overshoot of the last pass's counter:
+    weight -= 1 / 3;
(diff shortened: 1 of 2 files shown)


Iwasawafag:
It could be, that I broke upscaling with this. 
I'll fix it if needed and reopen this PR once I'll be sure upscaling works as good as it was

Iwasawafag:
Fixed. Lost incrementation of few indexes in two lines while merging simillar methods of the Resize class together.

As for the "bug" mentioned in Iwasawafag/jimp@65158a9 — it turned out to be some kind of a magic number, controlling how far image should shift vertically/horizontally. Changes introduced by this commit were reverted in master.

Iwasawafag:
I kept playing with that number (`1 / 3`) and noticed that the farther away original and target dimensions are from one another - the more noticeable image shifts to the top-left, especially if compared to output produced by Photoshop. 

I don't know if one third is a mathematically justified value or original programmer picked it at random, but I decided to make this value scale based on originalSize/targetSize proportion and I've got the results I'm happy with.

Example: (it's a 60x60 image resized to 512x512, then zoomed out to 25% in PS)

- [Nearest neighbor (resized in PS)](https://cloud.githubusercontent.com/assets/5304842/17722361/e86d7106-6431-11e6-91ea-9b533d5e85d9.png)
- [Resized with Jimp w/ original value](https://cloud.githubusercontent.com/assets/5304842/17722389/18011ada-6432-11e6-8b71-aec1fe22e8fa.png)
- [Resized with Jimp w/ adjusted value](https://cloud.githubusercontent.com/assets/5304842/17722394/24d6ce08-6432-11e6-93b1-17233ec0891a.png)

These changes could be found in Iwasawafag/jimp@8c27ae5 if you want them, but they probably should not be used, since it's not backed by any actual math and the difference in output isn't too big to begin with.

oliver-moran:
Merged. Thanks and sorry for the delay. Very busy IRL. Will do a publish shortly and include this.

Issue #188, opened 2016-11-02
I hope the image.write()  function support promise; 


Messilimeng:

hipstersmoothie:
close via #479

PR #189, opened 2016-11-05
#188 Add promise interface to 'write' method

doublemcz:


gutenye:
Any updates? I need it for `async/await` style.

index.js
```diff
@@ -2312,11 +2312,41 @@ function measureText(font, text) {
...
- * @param path a path to the destination file (either PNG or JPEG)
- * @param (optional) cb a function to call when the image is saved to disk
- * @returns this for chaining of methods
+ *
+ * @param {string} path a path to the destination file (either PNG or JPEG)
+ * @param {null|function} cb a function to call when the image is saved to disk
+ * @param {boolean} returnPromise
+ * @returns {Jimp|Promise} if returnPromise is true then returns a Promise otherwise `this` for chaining methods
+ */
+Jimp.prototype.write = function (path, cb, returnPromise) {
```
JakeChampion: The pattern I've seen in other codebases is to return a promise if the function was not called with a callback argument. The current approach in this PR requires that a callback argument has to be given, even if it is not being used.

doublemcz: It will be BC break. @oliver-moran Can you please decide as repo author? Thanks!

strandedcity: I think it makes sense to have a promise implementation for ```write()```, but I agree backwards compatibility is an issue, and it's strange to have to call Jimp with an empty callback when using the promise implementation.

In the case of Jimp, I believe this also breaks functionality when you want multiple outputs. Though this isn't how I've used Jimp myself, can't you do this for multiple thumbnail processing, and expect it to work?

```
.resize(1000,1000)
.write("/path/to/1000px-thumbnail")
.resize(500,500)
.write("/path/to/500px-thumbnail")
```

... at the end of the day, promises would be great, but the implementation here for ```.write()``` is problematic.

arackaf: @oliver-moran  asked me to help out with the PR backlog.  I'd say that since converting Node-style callback APIs into Promises is fairly straightforward for those who want it

http://exploringjs.com/es6/ch_promises.html#sec_nodejs-callbacks-promises

there's no real reason to break backward compat for it.  I'll close this out but of course if Oliver feels differently he can revive it :)

PR #192, opened 2016-11-12
Use .editorconfig

CodeLenny:
Adding a [EditorConfig](http://editorconfig.org/) file prevents editors from preforming bulk reformatting to match a different code style, as seen in f2842a7a8e6d407b05dc16e406f24ebba8f19040 on my fork (over 3k lines accidentally changed)

I've set both `trim_trailing_whitespace` to false, and `insert_final_newline` to true, as that's what the source seems to be using.  Alternatively, if you prefer not having trailing whitespace, you could take this opportunity to make the bulk change, and let EditorConfig enforce it for future changes.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,7 @@ indent_style = space
 indent_size = 4
 end_of_line = lf
 trim_trailing_whitespace = false
-insert_final_newline = false
+insert_final_newline = true
 
 [package.json]
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# See http://editorconfig.org/ for valid options and editor plugins.
+
+# Prevents EditorConfig from searching higher in the tree
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+trim_trailing_whitespace = false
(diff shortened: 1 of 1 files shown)


strandedcity:
@CodeLenny it appears that many of the bulk reformatting changes that occurred on your fork were in the ```browser/lib/jimp.js*``` files, which are the output of build processes.

Is there a way for ```.editorconfig``` to apply only to ```index.js``` ?

Iwasawafag:
Yes, but it I think it would be better to exclude entire `/browser/`directory instead. There are two ways of doing that:

Correct way would require creation of another `.editorconfig` file inside `/browser/` directory with only one line: 

```
root = true
```

That would prevent plugin from walking higher in directory tree and discovering another `.editorconfig` with all the rules. 

Another widely used method is to provide incorrect values for the scope that needs to be excluded. In which case plugin just falls back to default settings of the editor. It helps with having only one configuration file, though plugin may complain about those invalid values.

```
[/browser/**]
charset = none
end_of_line = none
indent_size = none
indent_style = none
insert_final_newline = none
trim_trailing_whitespace = none
```

strandedcity:
Yep, good solution. Excluding /browser sounds right to me.

PR #199, opened 2016-11-18
Add Text Alignment

Ratismal:
I have made two commits to implement text alignment.

Commit 1 - Text Changes
- No longer uses a recursive hell
- New `alignment` option - Jimp#print(font, x, y, text, maxWidth, alignment, cb)
- - `Jimp.ALIGN_FONT_LEFT`
- - `Jimp.ALIGN_FONT_CENTER`
- - `Jimp.ALIGN_FONT_RIGHT`
- Added mid-word text wrapping

Commit 2 - Minor Fixes
- Documentation
- Check if alignment is undefined
- Use incremental for loop instead of `for of` for compatibility

### DIFF
--- a/index.js
+++ b/index.js
@@ -2243,6 +2243,7 @@ function loadPages(dir, pages) {
  * @param y the y position to start drawing the text
  * @param text the text to draw
  * @param maxWidth (optional) the boundary width to draw in
+ * @param alignment (optional) the direction to align text
  * @param (optional) cb a function to call when the text is written
  * @returns this for chaining of methods
  */
@@ -2253,6 +2254,9 @@ Jimp.prototype.print = function(font, x, y, text, maxWidth, alignment, cb) {
     }
     if ("undefined" == typeof maxWidth) {
         maxWidth = Infinity;
+    }
+    if ("undefined" == typeof alignment) {
+        alignment = Jimp.ALIGN_FONT_LEFT;
     }
 
     if ("object" != typeof font)
@@ -2308,7 +2312,8 @@ Jimp.prototype.print = function(font, x, y, text, maxWidth, alignment, cb) {
         line: line,
         y: y
     });
-    for (var tempLine of lines) {
+    for (var i = 0; i < lines.length; i++) {
+        var tempLine = lines[i];
         var xOffset = 0;
         var text = tempLine.line.trim();
         if (alignment == 0) xOffset = (maxWidth - measureText(font, text)) / 2;
--- a/index.js
+++ b/index.js
@@ -394,6 +394,11 @@ Jimp.FONT_SANS_32_WHITE = Path.join(__dirname, "fonts/open-sans/open-sans-32-whi
 Jimp.FONT_SANS_64_WHITE = Path.join(__dirname, "fonts/open-sans/open-sans-64-white/open-sans-64-white.fnt");
 Jimp.FONT_SANS_128_WHITE = Path.join(__dirname, "fonts/open-sans/open-sans-128-white/open-sans-128-white.fnt");
 
+// Font alignments
+Jimp.ALIGN_FONT_LEFT = -1;
+Jimp.ALIGN_FONT_CENTER = 0;
+Jimp.ALIGN_FONT_RIGHT = 1;
+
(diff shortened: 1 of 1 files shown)


maqboolkhan:
this change is much needed and i am also expecting it in JIMP let see when it will merge :)

Iwasawafag:
This would be a breaking change for those, relying on old signature *and* using node-style callback parameter, wouldn't it? 

```js
image.print(font, x, y, str, maxWidth, function(err, img) {
    /* ... */
});
```

In this case callback is assigned to *alignment* and exception "alignment must be a number" is going to be thrown. Attribute reassignment logic should be updated to account for that case.

Iwasawafag:
Also there already are `HORIZONTAL_ALIGN_LEFT`, `HORIZONTAL_ALIGN_CENTER` and `HORIZONTAL_ALIGN_RIGHT`, why not reuse those constants? 

Notice how `Jimp.AUTO` is used in multiple situations (`resize`, `getBuffer`, `getBase64` methods)

index.js
```diff
@@ -2258,24 +2267,60 @@ Jimp.prototype.print = function (font, x, y, text, maxWidth, cb) {
         return throwError.call(this, "text must be a string", cb);
     if ("number" != typeof maxWidth)
         return throwError.call(this, "maxWidth must be a number", cb);
+    if ("number" != typeof maxWidth)
+        return throwError.call(this, "alignment must be a number", cb);
```
Iwasawafag: Probably forgot to change variable name in this line.  It's same as in the block above and judging from error message it should have been `alignment`

maqboolkhan:
@Iwasawafag can you kindly tell any progress on this?

masihyeganeh:
Sorry that I'm asking again, but is it possible to review this PR and merge it if it is ok?
If the only problem is backward compatibility, you can achieve this simply by something like:
```javascript
if ("undefined" == typeof cb && "function" == typeof alignment) {
    cb = alignment;
}
```

Let me tell you this. I'm working on a very important project that requires writing Persian text on image. Persian text are RTL and right-aligned. I somehow managed to do the RTL thing (I can make a PR to add RTL support for texts to JIMP if you are interested) but alignment can't be done from outside of project.

So if it is possible, please add this functionality.
Thanks

masihyeganeh:
It seems like this PR won't be merged soon. So anyone can add this to your `package.json` file to use this unmerged PR for now:
```json
{
  "dependencies": {
    "jimp": "git+https://github.com/oliver-moran/jimp.git#pull/199/head"
  }
}
```

hipstersmoothie:
Does anyone in this thread want to tell me how this is different than using the current constants?

Ratismal:
@hipstersmoothie not sure if any behaviour with the current constants has changed, considering that this PR is 2 years old, but iirc the main difference is text wrapping.

hipstersmoothie:
I think you can fix the breaking change by aligning with this syntax 

```js
Jimp.loadFont(path).then(function(font) {
    image.print(
        font,
        x,
        y,
        {
            text: 'Hello world!',
            alignmentX: Jimp.HORIZONTAL_ALIGN_CENTER,
            alignmentY: Jimp.VERTICAL_ALIGN_MIDDLE,
        },
        maxWidth,
        maxHeight
    ); // prints 'Hello world!' on an image, middle and center-aligned
});
```

unless of course the above can get you what you want. Could I see pictures showcasing the differences?

index.js
```diff
@@ -2258,24 +2267,60 @@ Jimp.prototype.print = function (font, x, y, text, maxWidth, cb) {
...
         var testLine = line + words[n] + ' ';
         var testWidth = measureText(font, testLine);
-        if (testWidth > maxWidth && n > 0) {
-            that = that.print(font, x, y, line);
-            line = words[n] + ' ';
-            y += font.common.lineHeight;
+        if (testWidth > maxWidth) {
+            if (measureText(font, line) > maxWidth) {
+                cutLine = '';
+                while (measureText(font, line + '-') > maxWidth && line.length > 1) {
```
hipstersmoothie: loops inside loops scare me

PR #203, opened 2016-11-27
Create from raw pixel data or older versions of Jimp

### LINKED ISSUE #200: Proposal: rely on duck typing to determine if input is a Jimp instance instead of comparing constructors
I think it would be great if we were able to instantiate Jimp image from another jimp-like object whose constructor is not necessary the same as constructor from the current instance.

And personally I think of one remarkable feature - `bitmap` object. It is a part of public API (lowlevel, but still public), was around for ever and unlikely to be changed in the foreseen future. It also contains all necessary data to make a copy of an image. So I would suggest adding a method that looks something like this

```js
Jimp.isJimp = function(image) {
    if (image && image.constructor === Jimp) return true;

    return image && image.bitmap 
        && typeof image.bitmap.width === 'number'
        && typeof image.bitmap.height === 'number'
        && Buffer.isBuffer(image.bitmap.data) 
        && image.bitmap.data.length === image.bitmap.width * image.bitmap.height * 4;
}
```

and using it in the constructor instead of current check `if ("object" == typeof arguments[0] && arguments[0].constructor == Jimp) {`

## Why?

So different modules could pass instances of Jimp around. Because otherwise we would have to encode image to something like PNG in one module just to decode it right away in another. Encoding/decoding takes time and wastes CPU cycles we don't want to waste.

It two or more modules in the project share same dependency – it would be rational thing to open/create and preprocess image in one module and send it as is to another in RAW format, if that other module known to accept Jimp instances as image sources. But the highter number of such packages used at the same time (think of something with modular architecture or a Gulp/Grunt-like toolchain) — the lower chances are that all of them will depend on the same version of Jimp.

Iwasawafag:
Allows creation of Jimp instances from older versions of the module (where `!img instanceof Jimp`) and from object with raw RGB/RGBA data. Closes #200

`new Buffer(...)` is deprecated. It may either show warnings in console or change it's behaviour to zerofilling. Discussion on what to do was going for some time at nodejs/node. "safe-buffer" module is used for backwards compatibility with node <= 0.12 (this extra dependency could be avoided, but calls to `Buffer.allocUnsafe` and `Buffer.from` looks nicer with it. Hopefully it's not a problem)

### DIFF
--- a/index.js
+++ b/index.js
@@ -585,6 +585,15 @@ Jimp.distance = function (img1, img2) {
     return phash.distance(hash1, hash2);
 }
 
+/**
+ * Checks if img is a Jimp image
+ * @static
+ * @param {Object} img
+ * @returns {Boolean}
+ */
+Jimp.isJimp = function (img) {
+    return isJimpStrict(img) || isJimpLoose(img);
+}
 
 // An object representing a bitmap in memory, comprising:
 //  - data: a buffer of the bitmap data
--- a/index.js
+++ b/index.js
@@ -16,6 +16,8 @@ var BigNumber = require('bignumber.js');
 var URLRegEx = require("url-regex");
 var BMFont = require("load-bmfont");
 var Path = require("path");
+var Buffer = require("safe-buffer").Buffer;
+var isBuffer = require("is-buffer");
 
 if (process.env.ENVIRONMENT !== "BROWSER") {
     //If we run into electron renderer process, use XHR method instead of Request node module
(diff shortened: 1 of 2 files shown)


Iwasawafag:
Hmm... Could additionally check if `data` is an instance of `Uint8ClampedArray`. This way it will also support [`ImageData`](https://developer.mozilla.org/docs/Web/API/ImageData) objects returned by [CanvasRenderingContext2D#getImageData](https://developer.mozilla.org/docs/Web/API/CanvasRenderingContext2D/getImageData). 

But that needs to come with the check that `Uint8ClampedArray` is supported, because in IE10 it isn't.

hipstersmoothie:
@Iwasawafag wanna update this?

hipstersmoothie:
closing in favor of #514

PR #204, opened 2016-11-28
Performance improvements

Iwasawafag:
Hello.

I noticed that some methods waste time by doing unnecessary work like [clonning newly created buffers](https://github.com/oliver-moran/jimp/blob/master/index.js#L1352-L1362) or [allocating buffers larger than needed](https://github.com/oliver-moran/jimp/blob/master/index.js#L838) and decided to fix that. So I changed that, added some shortcuts here and there and changed `simpleRotate` in a way so in doesnt have to iterate over every pixel more than once in for any angle supported.

Here are some numbers:

### `Jimp#rotate` (`simpleRotate`)

- Saves some time by not creating extra buffer. 
- Added special handling for 180 degree turn, rotation by 90/270 degrees always performed by the shortest route

```
// Rotate by 90 degrees
Jimp#rotate old x 230 ops/sec ±0.83% (189 runs sampled)
Jimp#rotate new x 257 ops/sec ±0.32% (196 runs sampled)

// Rotate by 180 degrees
Jimp#rotate old x 115 ops/sec ±0.48% (190 runs sampled)
Jimp#rotate new x 402 ops/sec ±0.17% (201 runs sampled)

// Rotate by 270 degrees
Jimp#rotate old x 77.28 ops/sec ±0.31% (176 runs sampled)
Jimp#rotate new x 253 ops/sec ±0.38% (196 runs sampled)
```

### `Jimp#flip`

- Saves some time by not creating extra buffer. 
- When flipping in both directions - uses simple rotation by 180 degree from above. 

```
// Flip horizontally
Jimp#rotate old x 31.34 ops/sec ±1.27% (169 runs sampled)
Jimp#rotate new x 92.31 ops/sec ±0.39% (181 runs sampled)

// Flip vertically
Jimp#rotate old x 31.10 ops/sec ±1.34% (168 runs sampled)
Jimp#rotate new x 102 ops/sec ±0.74% (186 runs sampled)

// Flip horizontally and vertically
Jimp#rotate old x 23.95 ops/sec ±1.11% (158 runs sampled)
Jimp#rotate new x 396 ops/sec ±0.49% (198 runs sampled)
```

### `Jimp#crop`

- Allocates buffer of correct length. 
- When possible - uses `Buffer#slice`, so cropped image would uses same memory as old buffer, just with different offset, 

(note: following numbers also include time of clonning test image in every run)

```
// Copy image and crop
Jimp#crop old x 252 ops/sec ±0.35% (193 runs sampled)
Jimp#crop new x 324 ops/sec ±0.76% (193 runs sampled)

// Copy image and crop without changing width
Jimp#crop old x 166 ops/sec ±0.63% (192 runs sampled)
Jimp#crop new x 1,199 ops/sec ±4.58% (179 runs sampled)
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,6 +16,8 @@ var BigNumber = require('bignumber.js');
 var URLRegEx = require("url-regex");
 var BMFont = require("load-bmfont");
 var Path = require("path");
+var Buffer = require("safe-buffer").Buffer;
+var isBuffer = require("is-buffer");
 
 if (process.env.ENVIRONMENT !== "BROWSER") {
     //If we run into electron renderer process, use XHR method instead of Request node module
@@ -126,7 +128,7 @@ function Jimp() {
             return throwError.call(this, "cb must be a function", cb);
 
         this.bitmap = {
-            data: new Buffer(w * h * 4),
+            data: Buffer.allocUnsafe(w * h * 4),
             width: w,
             height: h
         };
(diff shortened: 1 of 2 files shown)


Iwasawafag:
You may notice that code performing rotation by 90 degrees moved to a separate function. I had to do that, otherwise performance degraded seriously in benchmark after rotation by 180 degrees. Something is wrong with V8's optimizator, it seems. It was becoming even slower than old implementation, that performed 2 extra rotations

edi9999:
That's really great work !

hipstersmoothie:
closing in favor of #512

PR #185, opened 2016-10-31
Add convolution matrix method

aurium:
Sum neighbor pixels weighted by the kernel matrix. You can find a nice explanation with examples at [GIMP's Convolution Matrix plugin](https://docs.gimp.org/en/plug-in-convmatrix.html).

Closes #184

PR #218, opened 2016-12-28
Convolution and Displace filter, Pixelate effect.

maciejkrol:
- Convolution filter allows for applying any generic convolution filter to the image or a region.
- Pixelate effect with desired pixel size. Can be applied to the whole image or a region.
- Displacement filter moves pixels according to the provided displacement map (B&W image). This can be used to create a pair of images for 3D stereoscopic viewing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ image.blit( src, x, y[, srcx, srcy, srcw, srch] );
                                   // blit the image with another Jimp image at x, y, optionally cropped.
 image.composite( src, x, y );     // composites another Jimp image over this image at x, y
 image.mask( src, x, y );          // masks the image with another Jimp image at x, y using average pixel value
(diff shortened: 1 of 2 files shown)


strandedcity:
This is great, but can we separate the work for .displace() .pixelate() and .convolute()? 

There's another active PR for .convolute that works basically the same way (and also supports nxn kernels), but it does a nice job of handling the edge conditions as well: https://github.com/oliver-moran/jimp/pull/185

If we merge #185, I think we could also merge the .displace() and .pixelate functions here.

oliver-moran:
Thank you. I'll fixed up the duplicated function. Planning to do a release this weekend. This will be included.

oliver-moran:
I opted for this version @strandedcity because it supported regions. Would you mind adding the edge handling code to this?

PR #222, opened 2017-01-13
Add Type Definition file for TypeScript

russleyshaw:


strandedcity:
@russleyshaw any thoughts on how this file is maintained over time, assuming that not all contributors are using typescript? Already I see several PRs that would require changes here.

russleyshaw:
My initial `jimp.d.ts` isnt that large of a file. The changes needed to be made whenever code changes (such as #189) would be small in comparison to the functional changes.
Also, there is an ever-growing TypeScript community that would help contribute. Personally, one of the criteria I take into account when choosing which package to use is if it has typings.

Typescript users of Jimp would not need to include a `@types/jimp` dependency because the typings would be included in the repo. 

Yes, a `@types/jimp` could be created and maintained separately, but typically the preferred method is to include the types with the package itself.

aurium:
@russleyshaw do you know a good/simple way to test if the declared types are working? I'm thinking about TDD unity tests to ensure that declared types to still work on future releases (even if new methods do not be declared).

A good test may also help developers to know what to update at `jimp.d.ts`, when some method changes.

arackaf:
Merging this.  If the TS bindings get out of date, TS users can update / PR as needed (or hush the TS compiler with some `any` bindings), but it shouldn't be a breaking change for everyone else.

Issue #201, opened 2016-11-22
NPM package: --use_strict mode


ytiurin:
Hi Oliver,
When using Jimp in Node environment with the --use_strict mode enabled, dependency module `bmp-js` causes an exception. I raised an issue on `bmp-js` repo https://github.com/shaozilee/bmp-js/issues/8 and author fixed it in the latest version (0.0.3) on NPM.
Whould be great to have it up-to-date in Jimp package dependency.

Many production environments run in --use_strict mode so it would be great if Jimp also did.

Amazing lib!
Cheers :beer:

Tarpsvo:
Thanks for reporting it to `bmp-js` and for getting it fixed. I support this as I ran into the same problem. 👍

Dewep:
Same problem, it would be cool to update the package! :)

strandedcity:
https://github.com/oliver-moran/jimp/pull/225

PR #225, opened 2017-01-15
Upgrade bmp-js to 0.0.3

strandedcity:
Closes https://github.com/oliver-moran/jimp/issues/201

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "license": "MIT",
   "dependencies": {
     "bignumber.js": "^2.1.0",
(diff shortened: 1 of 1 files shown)


Iwasawafag:
Why don't we just merge it? The [difference between two versions](/shaozilee/bmp-js/compare/313795a...ed420af) is minimal, signatures of exported functions didn't change and I think it's safe to assume that nothing will break

oliver-moran:
Thank you. Planning to do a release this weekend. This will be included.

PR #228, opened 2017-01-20
Append image method added

dsilva2401:


Iwasawafag:
Looks like `Jimp#blit` method does the same thing and does it better, changing canvas size when needed

Iwasawafag:
Correction: `Jimp#blit` does not change size of canvas, but maybe it should. Then it will also cover functionality of ImageMagicks's `append` command.

Issue #214, opened 2016-12-15
New feature: Add option to disable auto-rotation


potwornaanakonda:
In some projects it would be very useful to disable auto-rotation based on EXIF metadata. E.g. when EXIF metadata is incorrect.

strandedcity:
Agreed, it would be useful to pass an ```options``` object to ```parseBitmap(data, mime, cb)```. ```enableExifRotate: true``` (as the default when options aren't specified) would be a great first option to have.

Any interest in coding it up?

hipstersmoothie:
proposed api:

```js
const image1 = Jimp.create({ exifRotate: true }) // off by default
const iamge2 = await Jimp.read(filePath, { exifRotate: true })
```

porfirialex:
Is there any work around to disable this?

Issue #21, opened 2015-06-04
Thank you


frankrousseau:
I just used your library, it works very well, it's super simple and solves a good problem. Thank you!

oliver-moran:
You're very welcome, Franks. Hope it's of help.

PR #245, opened 2017-03-02
Feature Request #214 : option to disable EXIF auto-rotation

kminkler:


geonanorch:
I am also interested in such a feature, because:
 - sometimes one would like to process pictures as-is
 - sometimes EXIF parsing seems buggy (see [issue 21](https://github.com/bwindels/exif-parser/issues/21) on the `exif-parser` project)

oliver-moran:
Thank you. Planning on doing a release this weekend. This will be included.

geonanorch:
Just tested: now images no longer automatically get rotated based on EXIF value (instead, `exifRotate()` must be called). Thank you!

geonanorch:
for the records, the [issue in the EXIF library (#21)](https://github.com/bwindels/exif-parser/issues/21) was fixed recently, so calling `exifRotate()` should be safer now.

PR #254, opened 2017-03-22
Raw constructor

jimkang:
Hi, I had a situation in which I needed to run Jimp on over 2000 jpegs. The profiler showed it was spending a lot of time in jpeg encoding/decoding, so I switched to using a C-based jpeg decoder. In order to get things back to Jimp to do further processing, I needed Jimp to be able to accept raw buffers (not jpeg buffers), so I added this constructor.


Iwasawafag:
There's another similar pull request: #203

hipstersmoothie:
closing in favor of #514

PR #257, opened 2017-04-07
Made methods more permissive when using subclasses of jimp

Haringat:
Currently various methods throw an error when people try to use a subclass of jimp instead of jimp itself. I don't actually see any benefit in forbidding to extend jimp so I made the changes.

### DIFF
--- a/index.js
+++ b/index.js
@@ -93,7 +93,7 @@ function throwError(error, cb) {
 /**
  * Jimp constructor (from another Jimp image)
  * @param image a Jimp image to clone
- * @param cb a function to call when the image is parsed to a bitmapf
+ * @param cb a function to call when the image is parsed to a bitmap
  */
 
 /**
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1061,12 +1061,12 @@ Jimp.prototype.autocrop = function() {
...
 */
 Jimp.prototype.blit = function (src, x, y, srcx, srcy, srcw, srch, cb) {
-    if ("object" != typeof src || src.constructor != Jimp)
+    if (!(src instanceof Jimp))
         return throwError.call(this, "The source must be a Jimp image", cb);
     if ("number" != typeof x || "number" != typeof y)
         return throwError.call(this, "x and y must be numbers", cb);
 
-    if ("function" == typeof srcx) {
+    if (srcx instanceof Function) {
```
borilla: Isn't it generally safer to use typeof to determine functions?

Haringat: I honestly couldn't think of a scenario where it makes a difference, but personally I would recommend to use typeof only for primitives.

Iwasawafag: The only case I can think of where `instanceof` check would fail is if `srcx` was defined in a different context (different frame or window). I think there's a chance of something like this happening when using browser version, but I've never used it myself and can't tell for sure.

I'd stick with `typeof` check personally

Iwasawafag:
Looks good to me, thanks!

PR #263, opened 2017-04-25
on write - create path if doesn't exist instead of failing silently

thewizarodofoz:


oliver-moran:
Thank you. Planing to do a release this weekend. This will be included.

PR #269, opened 2017-05-07
Fixed autocrop function, kept original reasoning

Willowblade:
Autocrop's eastPixelsToCrop were actually westPixelsToCrop, and vice versa, leading to images who are in the right half of the image to have a faulty crop (W x 1), due to the search from the south to the north yielding no results (wrong side of the image). This is fixed now.


oliver-moran:
Thank you. Planning to do a release this weekend. This will be included.

PR #275, opened 2017-05-21
Test refactor

aurium:


index.js
```diff
@@ -170,87 +150,116 @@ function Jimp() {
...
-            return throwError.call(this, "data must be a Buffer", cb);
-        if ("string" != typeof mime)
+        if (typeof mime !== "string")
             return throwError.call(this, "mime must be a string", cb);
-        if ("function" != typeof cb)
+        if (typeof cb !== "function")
             return throwError.call(this, "cb must be a function", cb);
 
-        parseBitmap.call(this, data, mime, cb);
+        parseBitmap.call(this, data, mime, cb, options);
```
Iwasawafag: I just noticed that Jimp constructor now accepts an `options` object. And it's only used in one place - here it's being passed to parseBitmap function...

index.js
```diff
@@ -271,30 +280,30 @@ function getMIMEFromPath(path, cb) {
...
+    return {
+        data: gifData,
+        width: gifObj.width,
+        height: gifObj.height
+    };
 }
 
 // parses a bitmap from the constructor to the JIMP bitmap property
-function parseBitmap(data, mime, cb) {
+function parseBitmap (data, mime, cb) {
```
Iwasawafag: ...and then it's completely ignored here.

Iwasawafag: I also checked and this function is not being replaced with anything else, so what's the purpose of that options object?
Can I remove it?

Iwasawafag: And it's not even used in tests either.

Iwasawafag:
Tests doesn't run in every environment either. On my Windows machine I'm facing two problems:

1. Symlink doesn't work. When I run tests it's being recognized as JS file by ESLint and fails linting. Why not make it regular js file and just put this inside?

```js
module.exports = require('./index');
```

2. Something goes wrong in `karma-chrome-launcher` module and I'm getting this error:

```
'CHROME_BIN' is not recognized as an internal or external command, operable program or batch file.
```

And I'm not sure if this is going to work with something like TravisCI, I have yet to find out.

I think you merged it way too early. What was the point of handling this through a merge request if you weren't going to wait for any kind feedback or discussion (for longer than 3 minutes)?

index.js
Iwasawafag: Never mind, I think I found it. It's sorta used in extra constructors call:

```js
new Promise(
    (resolve, reject)=> extraConstructor.run.call(this, resolve, reject, ...arguments)
)
```

index.js
```diff
@@ -170,87 +150,116 @@ function Jimp() {
...
+            cb = arguments[arguments.length-2]; // TODO: try to solve the args after cb problem.
+            if (typeof cb !== "function") cb = function () {};
+        }
+        var extraConstructor = Jimp.__extraConstructors.find((c)=> c.test(...arguments));
+        if (extraConstructor)
+            new Promise(
+                (resolve, reject)=> extraConstructor.run.call(this, resolve, reject, ...arguments)
+            )
+            .then(()=> cb.call(this, null, this))
+            .catch(cb);
```
Iwasawafag: What should probably be done instead is this:

```js
if (extraConstructor) {
    const possiblyAPromise = extraConstructor.run.call(this, ...arguments); // TODO: remove callback from arguments
    
    if (!possiblyAPromise || typeof possiblyAPromise.then !== 'function') {
        throw new TypeError('Incorrect return value of extra contructor, a Promise expected');
    }

    possiblyAPromise.then(
        ()=> setTimeout(()=> cb.call(this, null, this), 0),
        (err)=> setTimeout(()=> cb.call(this, err), 0)
    );
} else {
    /* ... */
```

This way any thrown error will remain uncaught and will crash the script and user will see his function at the top of the call stack

Iwasawafag:
I think I'm gonna give this a small rewrite. In face, already working on this

index.js
```diff
@@ -170,87 +150,116 @@ function Jimp() {
...
+        cb = arguments[arguments.length-1];
+        if (typeof cb !== "function") {
+            cb = arguments[arguments.length-2]; // TODO: try to solve the args after cb problem.
+            if (typeof cb !== "function") cb = function () {};
+        }
+        var extraConstructor = Jimp.__extraConstructors.find((c)=> c.test(...arguments));
+        if (extraConstructor)
+            new Promise(
+                (resolve, reject)=> extraConstructor.run.call(this, resolve, reject, ...arguments)
+            )
```
aurium: Sure! As i was solving so many things, i let this ugly working while its is not a publicized feature.

index.js
aurium: Good!

PR #285, opened 2017-05-31
Add ability to use alignment in print function

theill:
Extend print method to have `text` argument take either a string or an object. If an object it will be possible to specify an `alignment` parameter to align printed text

Similar to another PR but more backward compatible.


aurium:
Nice @theill! I'll be happy to accept your contribution. You only need to add tests to ensure your feature is working and it will still working on future. Please see [`print.test.js`](https://github.com/oliver-moran/jimp/blob/master/test/print.test.js)

theill:
@aurium sure. my bad. I've added tests now.

theill:
@aurium let me know if more changes are needed so we can get this merged into master

PR #288, opened 2017-06-16
Fix path bug for built-in fonts and delete symlink that doesn't work on Windows

danieltian:
This PR fixes the following 2 issues:

1. Fixed the path for built-in fonts. Previously it was `./`, which was getting the path from the base of the project that's using Jimp as a library, instead of from the Jimp folder in `node_modules`. Changed this back to `__dirname` and restored usage of `Path.join` to maintain Windows compatibility.

2. Removed the `jimp.js` symlink, as this is treated as a text file when cloned in Windows, and it was breaking the project. Modified `package.json` to point directly to `index.js`.


Iwasawafag:
I _think_ you could add a "browser" entry point in addition to "main" in `package.json` so browserify would look in the right place. Have literally zero experience with Browserify though, so I could be wrong

PR #294, opened 2017-06-27
Support IE11 in browser environment.

dianlight:
According to MSDN documentation the stack property is set to undefined when the error is constructed, and gets the trace information when the error is raised. 

Reference:  https://docs.microsoft.com/en-us/scripting/javascript/reference/stack-property-error-javascript


edi9999:
Thanks. It seems to me that the method should not be there at all, Creating an error just to know the location of the file seems a bit overengineered.

Can you rebase your MR onto master and regenerate the lib files ?

huytran0605:
@dianlight  when i tried to use your change, IE 11 throw error: `Access is denied`. Please help me to check :( i really love to use Jimp

hipstersmoothie:
I'd like to see this weird bit of code go too

hipstersmoothie:
closing in favor of #511

ezequielzacca:
Im still getting access is denied

PR #313, opened 2017-07-29
Fix a small rounding error in Nearest Neighbor algorithm

Longwelwind:
There seems to be a small error in the Nearest Neighbor algorithm for resizing images that produces unpleasant results for pixel-perfect images.

Below, I put a pixel-art image that I upscale by a factor of two. The first picture is the original pixel-art image, the second one is the upscaled version using the current algorithm, the third and last one is the upscaled version with the small modification I've made.

You can notice, for example, that the upper-right corner is not properly upscaled on the second screenshot.

paintdotnet_2017-07-29_11-39-57

twine_2017-07-29_11-36-18
twine_2017-07-29_11-38-58

### DIFF
--- a/resize2.js
+++ b/resize2.js
@@ -37,8 +37,8 @@ module.exports = {
             for (let j = 0; j < wDst; j++) {
                 var posDst = (i * wDst + j) * 4;
 
-                var iSrc = Math.round(i * hSrc / hDst);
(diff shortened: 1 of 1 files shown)


funcdoor:
This also fixes the problem of pixels "wrapping" to the other side on upscale. Absolutely needed fix for pixel art!

PR #314, opened 2017-07-29
intToRgba to intToRGBA in jimp.d.ts

from-nibly:


Ojas-Gulati:
I suggest you don't remove intToRgba - keep it and also define a new function intToRGBA in the main js file and make it point to intToRgba.

from-nibly:
intToRgba does not exist. is there a reason both cases should be supported? It will be more work and more confusing.

Ojas-Gulati:
Ah, ok. I thought that you were renaming an old feature.

Though something like this is on another feature of jimp - check `greyscale()` and `grayscale()` - they both point to the same function in the main js file.

patheticcockroach:
NB this is also fixed in this PR https://github.com/oliver-moran/jimp/pull/329 (which also covers a bunch of missing methods)

edi9999:
Fixed by #329

PR #315, opened 2017-07-31
Correct getPixelColor doc

GalHorowitz:
The @returns of getPixelColor is currently "the index of the pixel or -1 if not found" - probaly copied from getPixelIndex and forgot to change.


edi9999:
Thanks !

PR #329, opened 2017-08-28
Fix and extend jimp.d.ts

kolbma:
There are missing some functions and methods and the constructors mustn't return a value.

### DIFF
--- a/jimp.d.ts
+++ b/jimp.d.ts
@@ -46,6 +46,7 @@ declare namespace Jimp {
         mirror(horizontal: boolean, vertical: boolean, cb?: Jimp.ImageCallback): this;
         gaussian(r: number, cb?: Jimp.ImageCallback): this;
         blur(r: number, cb?: Jimp.ImageCallback): this;
+        convolution(kernel: any, edgeHandling: number | Jimp.ImageCallback, cb?: Jimp.ImageCallback): this;
 
         greyscale(cb?: Jimp.ImageCallback): this;
         grayscale(cb?: Jimp.ImageCallback): this;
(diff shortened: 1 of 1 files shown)


wenzheng:
Hi, any update for this pull request, it contains some fix that would be very beneficial for my project as well:-)

jimp.d.ts
```diff
@@ -46,22 +46,33 @@ declare namespace Jimp {
...
+        scale(f: number, mode?: string | Jimp.ImageCallback, cb?: Jimp.ImageCallback): this;
         scaleToFit(w: number, h: number, mode?: any, cb?: Jimp.ImageCallback): this;
-        rotate(deg: number, mode?: number|boolean, cb?: Jimp.ImageCallback): this;
+        pixelate(size: number, x: number, y: number, w: number, h: number, cb?: Jimp.ImageCallback): this;
+        convolute(kernel: any, x: number | Jimp.ImageCallback, y?: number, w?: number, h?: number, cb?: Jimp.ImageCallback): this;
+        rotate(deg: number, mode?: number | boolean, cb?: Jimp.ImageCallback): this;
+        displace(map: Jimp, offset: number, cb?: Jimp.ImageCallback): this;
         getBuffer(mime: string, cb:(err:Error, buffer:Buffer)=>any): this;
-
+        getBase64(mime: string, cb?: Jimp.ImageCallback): this;
```
cryptoquick: Actually, this is not an ImageCallback. It returns a Base64 string:
https://github.com/oliver-moran/jimp/blob/4a7b21994e8a209cac6d89c7fcb8f3f820d9919d/index.js#L2455
The line should be:
`getBase64(mime: string, cb?:(err:Error, src:string)=>any): this;`

kolbma: Thanks. See https://github.com/kolbma/jimp/commit/0acf115828bcd0686701b6a439f9adc0cd98c6e7

edi9999:
Thanks !

PR #348, opened 2017-09-20
Rewrite the type declarations file

VojtechStep:
Includes constants and correct return types for functions with callback.
To import Jimp correctly to your TypeScript project, import it like so: `import Jimp = require('jimp')`

### DIFF
--- a/jimp.d.ts
+++ b/jimp.d.ts
@@ -4,6 +4,12 @@ declare namespace Jimp {
     type ColorActionName = 'mix' | 'tint' | 'shade' | 'xor' | 'red' | 'green' | 'blue' | 'hue';
     type ColorAction = { apply: ColorActionName, params: any };
 
+    type PrintableText = string | {
+        text: string;
+        alignmentX: number;
+        alignmentY: number;
+    };
+
     interface Bitmap {
         data: Buffer;
         width: number;
@@ -19,6 +25,61 @@ declare namespace Jimp {
         g: number;
         b: number;
         a: number;
+    }
+
+    interface FontChar {
+        id: number;
+        x: number;
+        y: number;
+        width: number;
+        height: number;
+        xoffset: number;
+        yoffset: number;
+        xadvance: number;
+        page: number;
+        chnl: number;
+    }
+
+
+    interface FontInfo {
+        face: string;
+        size: number;
+        bold: number;
+        italic: number;
+        charset: string;
+        unicode: number;
+        stretchH: number;
+        smooth: number;
+        aa: number;
+        padding: [number, number, number, number];
+        spacing: [number, number];
+    }
+
+    interface FontCommon {
+        lineHeight: number;
+        base: number;
+        scaleW: number;
+        scaleH: number;
+        pages: number;
+        packed: number;
+        alphaChnl: number;
+        redChnl: number;
+        greenChnl: number;
+        blueChnl: number;
+    }
+
+    interface Font {
+        chars: {
+            [char: string]: FontChar;
+        };
+        kernings: {
+            [firstString: string]: {
+                [secondString: string]: number;
+            };
+        };
+        pages: string[];
+        common: FontCommon;
+        info: FontInfo;
     }
 
     class Jimp {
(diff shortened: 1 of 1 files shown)


chigix:
I'm looking for this pull request, because current declaration file would make visual code unable to recognize constructor.

edi9999:
Can you update your pull request @VojtechStep ?

There are some conflicts now that I merged other PR

edi9999:
There are still conflicts in your branch @atwoodjer , can you fix them ?

atwoodjer:
this is not my branch.

hipstersmoothie:
@VojtechStep update this and I'll merge quickly!

VojtechStep:
I'm going to take a look at it today

VojtechStep:
Don't merge yet please, I will submit the rest in the next 12 hours

jimp.d.ts
```diff
@@ -18,7 +41,120 @@ declare namespace Jimp {
...
+        static FONT_SANS_64_WHITE: string;
+        static FONT_SANS_128_WHITE: string;
+
+        // Edge Handling
+        static EDGE_EXTEND: 1;
+        static EDGE_WRAP: 2;
+        static EDGE_CROP: 3;
+
+        // Properties
+        bitmap: Bitmap;
```
hipstersmoothie: can you move this line back below the constructors? this apparently fixed something #393

jimp.d.ts
```diff
@@ -1,5 +1,28 @@
 declare namespace Jimp {
-    type ImageCallback = (err: Error | null, image: Jimp) => any;
+    type GenericCallback<T, U = any, TThis = any> = (
+        this: TThis,
```
hipstersmoothie: is `TThis` a mistake?

jimp.d.ts
```diff
@@ -30,285 +166,413 @@ declare namespace Jimp {
...
+        getWidth(): number;
+        inspect(): string;
+        toString(): string;
         getMIME(): string;
         getExtension(): string;
-        getPixelIndex(
+        write(path: string, cb?: Jimp.ImageCallback): this;
+        writeAsync(path: string): Promise<Jimp>;
+        deflateLevel(l: number): this;
+        deflateLevel<T>(l: number, cb: Jimp.ImageCallback<T>): T;
```
hipstersmoothie: Why does some of the have `<T>`? What is the purpose

jimp.d.ts
```diff
@@ -30,285 +166,413 @@ declare namespace Jimp {
...
+        deflateStrategy(s: number): this;
+        deflateStrategy<T>(s: number, cb: Jimp.ImageCallback<T>): T;
+        filterType(f: number): this;
+        filterType<T>(f: number, cb: Jimp.ImageCallback<T>): T;
+        rgba(bool: boolean): this;
+        rgba<T>(bool: boolean, cb: Jimp.ImageCallback<T>): T;
+        quality(n: number): this;
+        quality<T>(n: number, cb: Jimp.ImageCallback<T>): T;
+        getBase64(mime: string, cb?: GenericCallback<string, any, this>): this;
```
hipstersmoothie: Why is `GenericCallback` used in some places? They seem to have the same style callback as the rest

jimp.d.ts
```diff
@@ -30,285 +166,413 @@ declare namespace Jimp {
...
             h: number,
-            mode?: any,
-            cb?: Jimp.ImageCallback
-        ): this;
+            cb: Jimp.ImageCallback<T>
+        ): T;
+
+        // Color methods
+        brightness(val: number): this;
+        brightness<T>(val: number, cb: Jimp.ImageCallback<T>): T;
```
hipstersmoothie: is the repetition needed? why doesn't an optional param work?

jimp.d.ts
VojtechStep: It's not a mistake, it's a generic argument that specifies the value of `this` in the callback function

jimp.d.ts
VojtechStep: If you look at the code, at the end of those function there is an if statement to check whether you passed a callback. If not, the function returns the Jimp instance for chaining, otherwise it returns the result of the callback call. The generic argument is inferred from the return type of the callback

jimp.d.ts
VojtechStep: JimpCallback is just an alias for a GenericCallback with a Jimp instance as the passed value. I could've used `GenericCallback<Jimp, T, Jimp>` everywhere, but it was common enough to make it a separate type

jimp.d.ts
VojtechStep: Optional param doesn't work because the two versions have different return types

jimp.d.ts
VojtechStep: I'll take a look at it

VojtechStep: I created a typescript project using my forks version as a dependency and I can't reproduce the issue you mentioned, the issue was probably something else. Therefore moving the statics is a purely stylistic matter. Should I move it anyway?

VojtechStep:
This should be the final version, I've tested it with a TypeScript project and couldn't find any errors.

hipstersmoothie:
Thanks for your work @VojtechStep!

PR #352, opened 2017-09-30
Threshold Plugin

brownrw8:
Add "threshold" method to image. I believe this is useful as a simplified method for processing scanned drawings, signatures, etc.  I've used a couple other methods for this but I always end up having to chain multiple calls/set up variables to get the same effect.

```
var {Jimp, getTestDir} = require("./test-helper");

describe("Threshold", function () {

    this.timeout(15000);

    var originalRenditionPath = getTestDir() + "/samples/hands.jpg";
    var lighterRenditionPath = getTestDir() + "/samples/threshold-samples/hands_mx200_rp255.jpg";

    var originalRendition, lighterRendition;
    before((done)=> {
        var imgs = [];
        imgs.push(Jimp.read(originalRenditionPath));
        imgs.push(Jimp.read(lighterRenditionPath));
        Promise.all(imgs).then((loadedImgs)=> {
            originalRendition = loadedImgs[0];
            lighterRendition = loadedImgs[1];
            done();
        }).catch(done);
    });

    it("defines default threshold for lighter backgrounds", ()=> {
        var transformedOriginalRendition = originalRendition.threshold(200, 255);
        transformedOriginalRendition.hash().should.be.equal(lighterRendition.hash());
    });

});

```


edi9999:
Looks good to me. 

The only thing I don't like that much is the `autoGrayscale`

I find it better to let the users be more explicit. 

The fact that the API is chainable makes things easy too : 

image.grayScale(...).threshold(...)

So I suggest to remove the `autoGrayscale` parameter.

hipstersmoothie:
@edi9999 I have updated this PR and added `threshold` as a plugin. I think keeping the autogreyscale is fine since it helps get better output from scan.

edi9999:
Ok, fine

PR #361, opened 2017-10-22
Added an example for scan

xkisu:
Added a bit of code to show how to check if you're at the end of a scan, I found this useful for something I was doing and others might as well.


edi9999:
Thanks !

PR #382, opened 2018-01-04
Replace the usage of stream-to-buffer with raw-body

katzj:
stream-to-buffer is deprecated (https://www.npmjs.com/package/stream-to-buffer)
and raw-body from stream-utils is the pointed to replacement.


edi9999:
Thanks, all tests are passing !

PR #383, opened 2018-01-05
Fix convolution to new convolute name in README.md

ksjogo:


edi9999:
I'm not 100% sure if we really need both the `convolution` function, and the `convolute`, they seem to be doing almost the same thing, just that `convolution` doesn't handle to pass (x,y, w,h) for the region where to apply the convolution. 

But ok to merge this.

PR #386, opened 2018-01-11
Update README.md

bdsomer:
The current `README.md` file shows the use of Node.js-style callbacks in the `Jimp.read` function. Sadly, this is simply incorrect as of January 10th, 2018 in the current `master` branch (see that there is no callback argument [here](https://github.com/oliver-moran/jimp/blob/master/index.js#L325)).

An alternative solution would be to implement callbacks into the code, but I believe that moving towards promises will increase the longevity of this package and encourage other package maintainers/collaborators to do the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Jimp.read("http://www.example.com/path/to/lenna.jpg").then(function (image) {
 
 ### Basic methods ###
 
-Once the the promise is fulfilled, the following methods can be called on the image:
+Once the promise is fulfilled, the following methods can be called on the image:
 
 ```js
 /* Resize */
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,19 +49,23 @@ Jimp.read("./path/to/image.jpg").then(function (image) {
...
+}).catch(function (err) {
+    // handle an exception
 });
 ```
 
 
 ### Basic methods ###
 
-Once the callback is filed or the promise fulfilled, the following methods can be called on the image:
+Once the the promise is fulfilled, the following methods can be called on the image:
```
edi9999: typo here : the the => the

edi9999: ping

bdsomer:
@edi9999 updated. sorry for the late response.

edi9999:
Thanks !

PR #391, opened 2018-01-22
Return promise in readme example

BrianRosamilia:
In the Promise example you aren't actually returning the promise in the promise chain.  Hopefully this helps someone!


edi9999:
Thanks

PR #396, opened 2018-01-30
Fix - Allow for 0 threshold in Jimp.diff

manudwarf:
As you can see in [Pixelmatch](https://github.com/mapbox/pixelmatch/blob/master/index.js#L11):
```javascript
 var threshold = options.threshold === undefined ? 0.1 : options.threshold;
```

They allow for 0 threshold, which we need for some of our comparisons.

I've provided additional tests as well, feel free to point out if I missed anything :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -613,7 +613,7 @@ Jimp.diff = function (img1, img2, threshold) {
         }
     }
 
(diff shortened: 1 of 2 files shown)


edi9999:
Thanks !

PR #399, opened 2018-02-06
Add `leaveBorder` to AutoCrop

JR-Y:
Hi,

Added autocropnew with the capability of leaving a border un-cropped of the original image. Did not want to change the original function since I could not find a way easily to support legacy code.


hipstersmoothie:
@JR-Y can i see examples with images?

hipstersmoothie:
@edi9999 I have also updated this one. It relies on #461 being merged first though

PR #414, opened 2018-03-02
fix tiff support

jeremypatrickdahan:
Tiff support seems broken in the current version (f7b5e5b).
Code in this PR is tested and working.
`npm test` shows a lot of lint errors, but the style of the modification is in line with the existing code.
(Please be kind, it is my first pull request!)


index.js
```diff
@@ -398,11 +398,14 @@ function parseBitmap (data, path, cb) {
             }
 
         case Jimp.MIME_TIFF:
-            var tiff = (UTIF.decode(data)[0]);
+            var ifds = UTIF.decode(data)
+            var page = ifds[0];
+            UTIF.decodeImages(data, ifds);
```
hipstersmoothie: Is this line actually doing anything?

jeremypatrickdahan: Which line ? `var ifds = UTIF.decode(data)` ?

jeremypatrickdahan: I believe it modifies `ifds` in place.
I need to re-check this

hipstersmoothie:
I tested it out and it does in fact work. @jeremypatrickdahan can you update the TIFF test so it passes? I will merge once you do.

jeremypatrickdahan:
Ok, the original tiff image in the samples directory had strange values in the test (pixels coordinates outside of the image). I changed it to one from http://www.petitcolas.net/watermarking/image_database/ and changed the values accordingly, double checking the values manually in Photoshop.

The line `UTIF.decodeImages(data, ifds);` uses a function of utif.js only available in version 2.0.0, so I updated the package.json. All tests now pass.

hipstersmoothie:
Sweet! can you regenerate the lock file? Don't manually resolve the merge conflicts. You can just delete it and re-install.

hipstersmoothie:
@jeremypatrickdahan Big changes happened in the repo! to resolve your merge conflicts I would:

* remove package-lock.json and run `yarn` to add the new dep to the lock file
* regenerate the browser/lib/*.js files

Thanks for your work!

jeremypatrickdahan:
I'm a bit confused... I did run `yarn` but the resulting lockfile a syntax different from the `npm` one.
It did regenerate the browser/lib/*.js files, although I am not sure it did it correctly

PR #416, opened 2018-03-12
Remove `noAssert` argument

BridgeAR:
Support for the noAssert argument dropped in the upcoming Node.js
v.10. There should not be any implications by this. The input is from now on
always validated one way or the other.

Refs: nodejs/node#18395


BridgeAR:
Ping @oliver-moran

BridgeAR:
Ping @oliver-moran again

edi9999:
Looks good to me

Issue #177, opened 2016-10-17
unknown JPEG marker ffff


italoandrade:
I have an image that throw this error "unknown JPEG marker ffff", but the image is a normal JPEG. How can I fix this problem?

oliver-moran:
Thanks for this. Can you upload a test image?

italoandrade:
Yes, here it is (Please, don't mind the name or the text on the image)

![ai-cara-como-amo-a-zuera](https://cloud.githubusercontent.com/assets/19225266/19690150/741fd452-9aae-11e6-9d9e-2f1171b19a1c.jpeg)

alt234:
Seeing this issue as well. Any updates per chance?

dvv:
It's jpeg-js related -- https://github.com/eugeneware/jpeg-js/pull/32
May be to update the dep?

tim-king:
This is now fixed in jpeg-js 0.3.3 so update your dependency

rtwalz:
Still seeing this issue, is the dependency updated?

peter-mouland:
@rtwalz to see if the dependency has been updated, you can have a look in the `package.json`.

I guess, if this is causing any issues for anyone else, a PR would probably be appreciated and confirmation the update doesn't break anything else.

rtwalz:
@peter-mouland It was my fault, I was unintentionally removing the end from an image URL. Thanks for the quick response regardless!

carterbancroft:
This is not fixed in jpeg-js 0.3.3. After seeing the comment above I attempted to put together a PR but the unit tests still fail with unknown JPEG marker error.

It's fixed in this commit to jpeg-js https://github.com/eugeneware/jpeg-js/commit/e5de15abe710b607de9cf251d83c5a9b04b46d5c but it's not released yet. I asked in the original PR if we could get a release on that and if a new version is minted I'll submit a PR with tests for this issue.

PR #424, opened 2018-03-30
Use latest jpeg-js package that fixes "unknown jpeg marker" error

carterbancroft:
Fixes: https://github.com/oliver-moran/jimp/issues/177

When opening JPGs with fill bytes it will fail with the error "unknown JPEG marker ffff" even though it's a valid jpg. The issue has been fixed in the jpeg-js repo with this PR https://github.com/eugeneware/jpeg-js/pull/32.

These changes add no new linting issues but oiliver-moran/jimp:master seems to have linting issues on it's own, which I've ignored in this PR.


peter-mouland:
different versions of npm produce different lock files, maybe it's that.

PR #427, opened 2018-04-07
Extendable Jimp TypeScript interface

AtaS:
This PR allows Jimp TypeScript interface to be extended.

For example I promisifed the `getBuffer()` method as

```js
myjimp.prototype.getBufferAsync = util.promisify(jimp.prototype.getBuffer)
```

But without this PR I couldn't extend the interface because the entry point was a variable, rather than an interface. The variable's entry methods returned the interface.

With this PR I am able to do this easily:

```js
import jimp from 'jimp'
import util from 'util'

interface MyJimp extends Jimp.Jimp {
	getBufferAsync(mime: string): Promise<Buffer>
}

interface MyJimpEntry extends Jimp.JimpEntry<MyJimp> {
	prototype: MyJimp
}

const myjimp: MyJimpEntry = <MyJimpEntry> jimp

myjimp.prototype.getBufferAsync = util.promisify(jimp.prototype.getBuffer)

export default myjimp
```

And when I am chaining methods it still works and I am getting "MyJimp" extended interface rather than the original `Jimp` interface.

For instance the code block below works fine with this PR:

```js
var img = await jimp.read(body)
		const resizedBody = await img.cover(this.newSize, this.newSize)
			.quality(80)
			.getBufferAsync(this.mimeType);
```

### DIFF
--- a/jimp.d.ts
+++ b/jimp.d.ts
@@ -75,7 +75,7 @@ declare namespace Jimp {
         toString(): string;
     }
 
-    var Jimp: {
+    interface JimpEntry<T extends Jimp> {
 
         // used to auto resizing etc.
         AUTO: number;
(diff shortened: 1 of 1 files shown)


lygstate:
Can be closed in favor of #428

hipstersmoothie:
@AtaS does #428 solve your issue?

PR #429, opened 2018-04-12
Jimp.prototype.write: Option to provide/keep MIME type regardless of file extension

jagoe:
I use Jimp to resize uploaded images and want temporary files to have no extension in order to make maintenance and deletion easier. For that reason, I needed Jimp to be capable of accepting an optional MIME type parameter in the `write` method.

Use of the augmented method is documented.
Since there were no tests for the write method and I wasn't sure how to handle functional tests including the file system, I didn't write any.
I didn't update the version number.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,10 +227,10 @@ Online tools are also available to convert TTF fonts to BMFont format (e.g. [Lit
 
 ### Writing to files ###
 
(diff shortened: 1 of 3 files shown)


hipstersmoothie:
closing in favor of #403 also this is a breaking change. at the moment we don't plan on making any breaking changes

PR #430, opened 2018-04-12
Generate and ts constructor fix

### LINKED ISSUE #393: Creating an image compiles with an error in TypeScript
I'm trying to use Jimp with TypeScript in a following manner:

```TypeScript
import * as Jimp from "jimp";
let image = new Jimp(256, 256, function (err, image) {
});
```

It produces the following errors:

```
severity: 'Error'
message: 'Only a void function can be called with the 'new' keyword.'
at: '2,13'
source: 'ts'
code: '2350'

severity: 'Error'
message: ''new' expression, whose target lacks a construct signature, implicitly has an 'any' type.'
at: '2,13'
source: 'ts'
code: '7009'
```

However, when I ignore the TypeScript compile-time errors and debug the result, it works just as expected, i.e. the image is created successfully. So it looks like we have a problem with jimp.d.ts. In fact, it describes the corresponding constructor not as a constructor, but as a regular function.

My main hypothesis is that d.ts file was prepared for an older version of TypeScript and uses different ways to describe `Jimp` class (a function with a prototype). However, I'm new to TypeScript, so I may miss something.

I've tried playing around with module augmentation (using this technique for the first time, actually), but got a following error: `Cannot augment module 'jimp' because it resolves to a non-module entity.`

The project details are as follows.

* NodeJS environment.
* Jimp 0.2.28, installed with npm.
* TypeScript 2.6.2.
* `tsconfig.json`:
```JSON
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "sourceMap": true,
    "strict": true,
    "outDir": "dist",
    "lib": [
      "es5",
      "es2015.iterable"
    ]
  },
  "include": [
    "src/**/*"
  ],
  "exclude": [
    "node_modules",
    "**/*.spec.ts"
  ]
}
```

jagoe:
Some people (myself included) had some issues with TypeScript and the constructor to generate a new image.
Moving the constructors to the top of their class should fix that problem.

And since it fits the pattern for the other use cases of Jimp, I added a static `generate` function that creates a new image. That way it can be used the same as, say, read to create a Jimp instance.

fixes #393

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,6 +436,18 @@ var image = new Jimp(256, 256, 0xFF0000FF, function (err, image) {
 });
 ```
 
+The same can be achieved by calling the static `generate` function:
+
+```js
+var image = Jimp.generate(256, 256, function(err, image) {
+    // this image is 256 x 256, every pixel is set to 0x00000000
+});
+
+var red = Jimp.generate(256, 256, 0xFF0000FF, function(err, image) {
+    // this image is 256 x 256, every pixel is set to 0xFF0000FF
+});
+```
+
 ## Comparing images ##
 
 To generate a [perceptual hash](https://en.wikipedia.org/wiki/Perceptual_hashing) of a Jimp image, based on the [pHash](http://phash.org/) algorithm, use:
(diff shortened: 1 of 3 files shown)


hipstersmoothie:
I don't think the generate function should really be in this PR. If you could move it to another PR that'd be awesome. 

Will merge the typescript changes when you resolve the merge conflicts!

hipstersmoothie:
closing in favor of #513

PR #460, opened 2018-07-12
promisify getBuffer

hipstersmoothie:


hipstersmoothie:
I think we can be backwards compatible pretty easy. Do you think we should
support both node callback style and promise? Or just promises?

Issue #223, opened 2017-01-14
Community Maintainer Guidelines


strandedcity:
If we're going to start having more people involved in Jimp's development, it probably behooves the community to have clear answers for common questions like:
- If I have a question, what do I need to do to get an answer?
- When can an issue / PR be closed?
- When making pull requests, what are code style and syntax guidelines to follow (there are several active issues about ES2015 / typescript / ```.editorconfig``` / coffeescript right now)
- What are the testing requirements for PR's to be merged
- Automated testing?
- Jimp conventions & clear development priorities
- Running list of bugs to be fixed and desired features for development

Additionally I think some conventions might be useful. For example, I think we're starting to have a need for "optional" parameters to be passed into some Jimp function, and other Jimp functions which accept so many parameters that having them in a specified order becomes confusing. Mrdoob of threejs draws the line at three, but the terrific work in https://github.com/oliver-moran/jimp/pull/191/files implements a function with seven:
```
Jimp.prototype.composite = function (src, x, y, mode, opacitySource, opacityDest, cb)
```
Should that be this instead?
```
Jimp.prototype.composite = function (src, options, cb)
```
where ```options``` is an object with defaults like:
```
{
   x: 0,
   y: 0,
   mode: Jimp.BLEND_SOURCE_OVER,
   opacitySource: 1,
   opacityDest: 1
}
```
?

All just food for thought.

aurium:
_Automated testing?_ Sure! On my ToDo to start and mandatory to all new PRs.

I like that! `Jimp.prototype.composite = function (src, options, cb)`

Iwasawafag:
If we're gonna have options objects we should probably provide extensive information for autocompletion engines. Does anyone know how many of those can use `.d.ts` files as information sources? 

Maybe we wont even need documentation comments anymore. But in case we still do - we should probably normalize those and mention that in contributor guidelines. Right now most of pull requests comes with proper JSDoc3 annotations, while existing documentation comments written in some other format, which I don't recognize, and they don't provide any type information.

I cant rewrite those into proper-ish JSDoc comments.

iamropo:
We should follow  a standard JavaScript style guideline to avoid conflicts, preferably [http://standardjs.com/ ](url)

Iwasawafag:
> "standard" style guideline

No thanks.

Iwasawafag:
[...]

PR #461, opened 2018-07-13
Add 'cropSymmetric' to AutoCrop

f-barth:
Why?
* I have a use-case that needs autocropping, but in a symmetric way. That means if I crop 2 pixels from the left I also want to crop 2 pixels from the right to keep the image centred. 

What?
* Add configuration of autocrop via options object: The parameters to autocrop were parsed "by-type" which did not allow adding a second boolean flag. I added the option to configure autocrop via an options object. 
* Add symmetric autocrop: if enabled "horizontal crop = Math.min(left crop, right crop)", same for vertical
* Add tests for the two above.

Notes:
* #223 mentioned the idea of using option objects instead of maaany parameters. I took up the idea. `autocrop` still is does handle the "old" invocation via parameters. All "old" parameters and symmetric autocropping can also be configured via options object.


hipstersmoothie:
@edi9999 This one is good to go now too.

PR #467, opened 2018-07-22
add omggif as dependency rather than file

hipstersmoothie:


hipstersmoothie:
@edi9999 any comments?

edi9999:
Looks good to me, but has conflicts now.

hipstersmoothie:
@edi9999 

* fixed merge conflicts
* noticed that `index.js` was indented wrong and i think i was because of clashing tab widths in prettier and editorconfig. so i set prettier's tab width to 4 to match and reformatted `index.js`

edi9999:
Yes, looks good to me !

PR #469, opened 2018-07-23
remove package lock

hipstersmoothie:
Since we now have a `yarn.lock` in the repo I think we should remove the `package-lock.json`. I have run into problems using the `package-lock.json` where with the `yarn.lock` I have never encountered a problem. Having only 1 lock file makes sense anyway.

@edi9999 ?


edi9999:
Yes, makes sense !

PR #470, opened 2018-07-23
upgrade dependancies

hipstersmoothie:
* uglifyjs updated API. much simpler
* bignumber needs full 64 alphabet set
* mimetype changed method names
* bmp.js seems to be doing thing more efficiently (just a guess)

Tried loading the and writing the test image and it seemed to work. The resulting image is indeed smaller than the source image.

@edi9999


hipstersmoothie:
@edi9999 do you mind if i switch the indentation to 2 spaces instead of 4? prettier breaks up the code a lot less on 2 spaces.

4 spaces:
screen shot 2018-07-23 at 1 17 41 am

2 spaces:
screen shot 2018-07-23 at 1 17 56 am

PR #464, opened 2018-07-21
Format Code with Prettier and use Stricter Linter

hipstersmoothie:
* Add prettier for consistent code formatting. Make contributing and reading the code a lot easier
	* took out editor config in favor of prettier
* Switch from eslint to XO. provides a far stricter linter and conforms better to modern JS practices. 
* Add babel-preset-env to get ES features. I haven't actually modified the code to be to ESexy yet. That'll come in another PR

@edi9999 Would you like to review?

PR #472, opened 2018-07-23
Make the promise polyfill a dev dependency that's only used to build the browser version

realityking:
Less dependencies are always nicer :)

Since #464 was merged it looks like Node.js 4 will be required anyway (use of let/const) so this does not affect which Node.js versions jimp will work with.


hipstersmoothie:
I was also doing this! Does #473 solve this? I think babel-preset-env should do all the heavy lifting.

realityking:
Hah, nice. :)

I'll keep this open in case #473 doesn't make it but yes, only one needs to go in.

Issue #253, opened 2017-03-18
Which files need to be hosted?


sufuf3:
Hi @oliver-moran  , 
I'm a member of [cdnjs](https://github.com/cdnjs/cdnjs). We are going to host this library.

I know that after v0.2.22, we should add the files under `browser/lib`.
But before v0.2.21 include v0.2.21, what the files need to be added in cdnjs?

If you have any suggestions, please let me know.

Thank you.

cdnjs/cdnjs#10855

oliver-moran:
Hi, @sufuf3. The only file needing hosting is https://github.com/oliver-moran/jimp/blob/master/browser/lib/jimp.min.js. Unfortunately, the latest build is broken. I'll ping you when all is working again.

extend1994:
Hi @oliver-moran I saw there is 0.2.28 tag since four month ago. Was the problem solved? If it was solved, we can add this to CDNJS soon 😄

hipstersmoothie:
@extend1994 @sufuf3 we now have a working version on npm 💯

we have:

* a browserified version under `browser`
* a babeled to es5 version under `main` in `dist`
* and esmodules version under `es`

extend1994:
@hipstersmoothie Thanks a lot! 👍

sufuf3:
Thank you.

PR #473, opened 2018-07-23
ES6

### LINKED ISSUE #75: Browserified version

### LINKED ISSUE #274: Browser version broken in 0.2.28
@strandedcity, dammit the browser version has broken again in this release. Any ideas what is up with it? I know too it is being used by many projects so this is a bit of a catastrophe.

### LINKED ISSUE #276: Node version support
@aurium great work with all of those tests and getting the browser version working again. Thank you.

One important question: What version of node should we support? Features like `let` and destructuring assignment are supported in Node 4.x, which is still in maintenance support and used on a lot of servers: https://github.com/nodejs/LTS#lts-schedule1

Any suggestion on the best way to go about moving forward with this? Perhaps Babel? https://babeljs.io/

### LINKED ISSUE #466: Jimp attached to window
When using the module in a webpack build, Jimp is exposed via the window rather than the `require` method just working. I would expect the window to not be used at all and for the following to work

```js
var Jimp = require("jimp");

var url = "https://upload.wikimedia.org/wikipedia/commons/0/01/Bot-Test.jpg";

Jimp.read(url).then(function (image) {
    image.greyscale().getBuffer(window.Jimp.MIME_JPEG, onBuffer);
}).catch(function (err) {
    console.error(err);
})

function onBuffer (err, buffer) {
  if (err) throw err;
  console.log(buffer);
}
```

currently you have to do the following to use Jimp in the browser:

```js
require("jimp");

var url = "https://upload.wikimedia.org/wikipedia/commons/0/01/Bot-Test.jpg";

window.Jimp.read(url).then(function (image) {
    image.greyscale().getBuffer(window.Jimp.MIME_JPEG, onBuffer);
}).catch(function (err) {
    console.error(err);
})

function onBuffer (err, buffer) {
  if (err) throw err;
  console.log(buffer);
}
```

hipstersmoothie:
* Put all src files in `src`. was kinda messy having everything at root.
* transpile code with babel for node builds. code is placed in the `dist` folder
	* main now points at the `dist` folder
* ignore all built files such as files in `browser/lib` and `dist`
* Babel all the files with `babel-preset-env`
	* remove the need for Promise polyfill #472 
	* remove `babel-preset-es2015` covered by `env`
* Build no longer runs after install
* removed `test:node:once:nocover` since it was doing the same thing as `test:node:run`
* started factoring out modules. Only logging and error-checking 

closes #466 
closes #75
closes #276 - with babel transpiling the node version we support many versions of node
closes #274

@edi9999


src/index.js
```diff
@@ -1,92 +1,44 @@
...
-const EventEmitter = require('events');
-
-const { PNG } = require('pngjs');
-const JPEG = require('jpeg-js');
-const BMP = require('bmp-js');
-const GIF = require('omggif');
-const UTIF = require('utif');
-const MIME = require('mime');
-const tinyColor = require('tinycolor2');
-const BigNumber = require('bignumber.js');
```
edi9999: I'm not in favor of using import syntax as it is not yet supported by node.

I would like to keep node-compatibility for the modules without any transpilation.

hipstersmoothie: we already babel so why not transpile for node as well?

hipstersmoothie: it's literally as easy as pointing to `dist/index.js` rather than `src/index.js`

hipstersmoothie: With this change we can also start supporting the `module` field in the package.json

https://github.com/nodejs/node-eps/blob/4217dca299d89c8c18ac44c878b5fe9581974ef3/002-es6-modules.md#51-determining-if-source-is-an-es-module

hipstersmoothie: a tree-shakable jimp is a goal of mine.

hipstersmoothie: https://github.com/webpack/webpack/issues/4674

hipstersmoothie: we already use a bunch of other es6 features as well that we should be transpiling for older platforms (const, let). why not import/export?

hipstersmoothie: After a bit of further investigation it seems like tree-shaking/elimination of unused class/prototype methods isn't really possible.

I'm still in favor of using import/export but am interested in reason why not to? We use es6 features not supported for node <6.4.0 so transpiling should already be happening.

hipstersmoothie: Although if they were to use the closure compiler unused prototype methods seem to be eliminated.

https://closure-compiler.appspot.com/home#code%3Dclass%2520Foo%2520extends%2520EventEmitter%2520%257B%250A%2509toString()%2520%257B%250A%2509%2509return%2520'fooooo'%253B%250A%2520%2520%2520%2520%257D%250A%257D%250A%250Aconst%2520toBeEliminated%2520%253D%2520()%2520%253D%253E%2520'be%2520gone!'%253B%250AFoo.prototype.toBeEliminated%2520%253D%2520toBeEliminated%250A%250Aconsole.log('here')%250Aconst%2520foo%2520%253D%2520new%2520Foo()%253B%250Aconsole.log(Foo.toString())

hipstersmoothie:
How would you feel about using https://github.com/developit/microbundle/blob/master/README.md ? Would could support many different ecosystems all it once!

hipstersmoothie:
We could also use https://www.npmjs.com/package/esm to output a ES Modules build.

So in total we could have:

* main - babelified build
* browser - browser build
* module - ES Modules build

hipstersmoothie:
@edi9999 what do you think? do you only want to support a non babelified cjs main and a browser field?

Cons:
* Support for node < 6.4 dropped (no es6 before that version)
* No support for module field (the future)

Pros:
* simple

edi9999:
hello hipster, for the import export, I'm also ok to use it, what I would however like to have is to compile the files without losing file/numbers, so basically only transpile what is really needed, because in node.js, you usually don't use sourcemaps.

edi9999:
The other problem that I see with using ES6 imports (and transpilation in general), is that when you have a stacktrace, you will have to change the path if you use an IDE or tool to "jump" to that position.

src/index.js
```diff
@@ -1,92 +1,44 @@
...
-const bMFont = require('load-bmfont');
-const MkDirP = require('mkdirp');
-const rawBody = require('raw-body');
-const fileType = require('file-type');
-const pixelMatch = require('pixelmatch');
-const EXIFParser = require('exif-parser');
-// polyfill Promise for Node < 0.12
-const Promise = global.Promise || require('es6-promise').Promise;
```
realityking: Removing this means you can drop es6-promise from the dependencies.

hipstersmoothie:
I just tested this and my IDE jumped to the correct area with source maps on. I followed this https://schempy.com/2015/05/20/source_maps_with_nodejs/. The source maps seems to work. @edi9999  Would you like to try it out?

src/index.js
hipstersmoothie: 👍

src/index.js
```diff
@@ -1,92 +1,45 @@
...
+import MIME from 'mime';
+import tinyColor from 'tinycolor2';
+import BigNumber from 'bignumber.js';
+import bMFont from 'load-bmfont';
+import MkDirP from 'mkdirp';
+import rawBody from 'raw-body';
+import fileType from 'file-type';
+import pixelMatch from 'pixelmatch';
+import EXIFParser from 'exif-parser';
+import sourceMaps from 'source-map-support';
```
edi9999: This should not be in the src file, since this means that we force users to have this dependency for sourcemaps, but for the browser usecase, it is already handled by the browser.

hipstersmoothie: This npm module works with browserify. The readme alludes to it.

hipstersmoothie:
@edi9999 I dont think that is really gonna be possible. We don't have optional source maps for browser, so why make them optional for node?

hipstersmoothie:
I could optionally require it based on the NODE_ENV. And try to set up a debug build. But I don't know where i would put the debug build in the package JSON. Or maybe if NODE_ENV is set to prod, dont use the source maps. 

What are the requirements you need me to satisfy for this to be merged?

hipstersmoothie:
@edi9999 the last commit ensures that production builds of jimp don't include source maps. but you can run `npm run build:node:debug` to get source maps.

hipstersmoothie:
@edi9999 I just did something pretty big here (I can move it to another branch if you want). Commit 16b2d94 fixes the browser build!!

The following now actaully works! 

```js
var Jimp = require('jimp');

Jimp.read('./exampleImages/dice.png')
  .then(...);

```

hipstersmoothie:
@edi9999 have now verified that source maps work in tests also. Source maps are only included when `BABEL_ENV` is set to `development`

hipstersmoothie:
@edi9999 after this i have a branch that breaks up the huge `index.js` file. Make moving through the project so much easier 👍

hipstersmoothie:
@edi9999 :) if more source maps issues come up I'm committed to fixing them.

hipstersmoothie:
@edi9999 I wanna get this released soon. I've been buggin @oliver-moran on twitter about it. This merge will fix both the browser version and use es5 for older versions of node via babel. 

Then we can contact the fine folks at cdnjs to start hosting a version of jimp that works on all platforms! #253

It's been almost a year since jimp worked in browser. Let's make it happen!

hipstersmoothie:
@oliver-moran can we merge this? I'd really like this in for the 0.3.0 release. 

as well as #477 and #486

PR #474, opened 2018-07-24
exact range is better

hipstersmoothie:
https://blog.kentcdodds.com/why-semver-ranges-are-literally-the-worst-817cdcb09277

and also the tool I use didn't pick up a few dependancies.

I was able to remove raw body by using the png.js sync api. I had to edit a test, but the output is smaller. Comparing the base64 output in a visualizer nothing had changed.

Removing raw-body results in ~0.3mb being removed from the final output.


hipstersmoothie:
@edi9999

hipstersmoothie:
@edi9999 I removed the carets. this PR now just updates dependancies a little more and removes raw-body

PR #479, opened 2018-07-26
Promises

hipstersmoothie:
Proof of concept 

enables the following syntax.

```js
const image = new Jimp('./exampleImages/dice.png')
  .scale(0.5)
  .grayscale()
  .write('image-small.png')
  .blur(50)
  .write('image-blur.png')
  .then(res => {
    console.log('all done', res);
  });
```

* all prototype methods are turned into promises
* `then` must come at the end of the chained methods
* can pass 'sync' as the last parameter to `new Jimp(...options, 'sync')` and the constructor and methods act synchronously. 
	* still doesn't work for filePaths or custom constructors

Here is where all the magic happens:

```js
	jimpPromise = cb => {
        const p = new Promise(cb);

		// Wrap each jimp method in a jimpPromise. This enables the above syntax
        Object.entries(this)
            .filter(([name]) => name[0] !== '_')
            .forEach(([method, func]) => {
                p[method] = (...args) =>
                    this.jimpPromise(resolve =>
                        p.then(image => resolve(this[method](...args)))
                    );
            });

		// Expose the normal EventEmiiter `on` function 
        p.on = (...args) => this.on(...args);

        return p;
    };
```

- [ ] change get buffer back to still return this


hipstersmoothie:
this would be a huge breaking change though. And it leads to weird states weird you want a sync jimp and would have to return promise in new places. I'm gonna tinker with this one a bit more.

PR #477, opened 2018-07-25
Break up Project into modules

hipstersmoothie:
relies on #473 being merged and closes #476 

* Constants
	* holds all `jimp` contstants
	* attached directly to `jimp` class (`Jimp[constant] = value;`)
* Image Manipulation functions
	* moved all similar image manipulation functions into separate files
	* color - (hue, saturation, sepia, ...)
	* shape - (rotate, crop, contain, ...)
	* effects - (blur, normalize, dither, ...)
	* text - just print for now. but other functionality that inserts text onto images can live here
* `image-bitmap.js`
	* contains `parseBitmap` and `getBuffer`
	* these two functions interact with image library dependancies (png-js, jpeg-js, bmp-js) so it made sense to group them together

PR #486, opened 2018-07-27
Promisify Functions

### LINKED ISSUE #90: Promise based write/getBuffer
Currently, write needs callback.
I'd like to be able to do this:

```
image.getBuffer(Jimp.MIME_JPEG).then(function(buffer) {
  res.writeHead(200, {
     'Content-Type': Jimp.MIME_JPEG,
     'Content-Length': buffer.length
   });
   res.end(buffer);
})
```

### LINKED ISSUE #232: The write promise callback is called before file exists
I just noticed that when using the promise based API, at least the write method actually resolves before the file was created, which is probably not what one would expect.

### LINKED ISSUE #308: Jimp.create instead of new Jimp for clearer syntax [Feature request] [Breaking change]
Currently, in order to create new image, one need to do:
```
var newImage = new Jimp(256, 256, function (err, image) {
    // this image is 256 x 256, every pixel is set to 0x00000000 
});
```
but this code is horrible, the `newImage` itself is undefined forever, not even a promise. I propose to change to following:
```
Jimp.create(256, 256),then((img) => {
    img.write(path);
    // this image is 256 x 256, every pixel is set to 0x00000000
}).catch(err => console.dir(err));
```

Update:

I think sync version can still stay this way:
```
var newImage = new Jimp(256, 256);
// this image is 256 x 256, every pixel is set to 0x00000000 
```
but async version should be deprecated and removed from doc as the reason I described above.

### LINKED ISSUE #374: write, getBuffer, getBase64 can be Promised
When using `async/await`, it's good way to have `write, getBuffer, getBase64` to return a Promise in addition to callback.

Now I work around this:

```js
  await jimp.read(imagePath)
    .then(image=>image.rotate(90))
    .then(image=>new Promise(
      (res, rej)=>image.write(savePath, (err,ret)=> err ? rej(err) : res(ret))
    ))
```

### LINKED ISSUE #445: new Jimp as promise
I see quite a bit of chatter about promises across various issues on the repo but no where do I see a mention about making the `new Jimp()` work with promises so I'm creating an issue addressing that. Currently the only way to create a new Jimp image appears to be as follows, adopted from the docs:

```js
new Jimp(256, 256, function (err, image) {
  if(!err) {
    // this image is 256 x 256, every pixel is set to 0x00000000
    return console.log(image);
  }
  return console.error(err);
});
```

but with promises this could be enhanced to:

```js
new Jimp(256, 256).then((img) => {
  // this image is 256 x 256, every pixel is set to 0x00000000
  return console.log(image);
}).catch((err) => {
  return console.error(err);
});
```

This currently throws the error message `"(intermediate value).then is not a function"`.
***

Also an example when using async/await

```js
asyc run () {
  try {
    var image = await new Jimp(256, 256);
    return console.log(image);
  } catch (err) {
    return console.error(err);
  }
}
```

hipstersmoothie:
closes #232 - write now can return a promise
closes #90 - same as above and getBuffer can now be async
closes #374 - base64 can now be async
closes #308 closes #445 - Jimp.read can take all args that the constructor can. Jimp.create added for clearer syntax. No need to add promises to the jimp constructor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,7 @@ The image can be written to disk in PNG, JPEG or BMP format (determined by the f
 
 ```js
 image.write(path, cb); // Node-style callback will be fired when write is successful
+image.writeAsync(path); // Returns Promise
 ```
 
 The original extension for an image (or "png") can accessed as using `image.getExtension()`. The following will save an image using its original format:
@@ -286,6 +287,7 @@ A PNG, JPEG or BMP binary Buffer of an image (e.g. for storage in a database) ca
 
 ```js
 image.getBuffer(mime, cb); // Node-style callback will be fired with result
+image.getBufferAsync(mime); // Returns Promise
 ```
 
 For convenience, supported MIME types are available as static properties:
@@ -304,6 +306,7 @@ A Base64 data URI can be generated in the same way as a Buffer, using:
 
 ```js
 image.getBase64(mime, cb); // Node-style callback will be fired with result
+image.getBase64Async(mime); // Returns Promise
 ```
 
 ### PNG and JPEG quality
--- a/jimp.d.ts
+++ b/jimp.d.ts
@@ -181,12 +181,15 @@ declare namespace Jimp {
         exifRotate(): this;
         displace(map: Jimp, offset: number, cb?: Jimp.ImageCallback): this;
         getBuffer(mime: string, cb: (err: Error, buffer: Buffer) => any): this;
+        getBufferAsync(mime: string): Promise<Jimp>;
         getBase64(mime: string, cb?: (err: Error, src: string) => any): this;
+        getBase64Async(mime: string): Promise<Jimp>;
         dither565(cb?: Jimp.ImageCallback): this;
         dither16(cb?: Jimp.ImageCallback): this;
         color(actions: any, cb?: Jimp.ImageCallback): this;
         colour(actions: any, cb?: Jimp.ImageCallback): this;
         write(path: string, cb?: Jimp.ImageCallback): this;
+        writeAsync(path: string): Promise<Jimp>;
         print(
             font: any,
             x: number,
--- a/src/index.js
+++ b/src/index.js
@@ -18,7 +18,7 @@ import * as effects from './image-manipulation/effects';
 
 import isDef from './utils/is-def';
 import { clear } from './utils/log';
-import { parseBitmap, getBuffer } from './utils/image-bitmap';
+import { parseBitmap, getBuffer, getBufferAsync } from './utils/image-bitmap';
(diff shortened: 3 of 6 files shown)


hipstersmoothie:
@oliver-moran If you just merge this one all #473 and #477 will be merged as well

hipstersmoothie:
@oliver-moran and if this one gets merged 10 issues get closed all at once 😮

hipstersmoothie:
I think I'm gonna change it from getBuffer(mime, 'async') to getBufferAsync(mime)

hipstersmoothie:
@oliver-moran Just added the 'async' functions. They are tested and documented. Everythings runs fine. This should be ready to go!

hipstersmoothie:
@edi9999 what do you think of this one? it gets makes of the requested functions promises without breaking the current api

edi9999:
in this MR I find that the duplication in the code is not good for maintanability.

Things like Bluebird can do the automatic transformation from callback to Promises.

hipstersmoothie:
So should I use bluebird to do this? Or you want users to do that? I think it would be better just to have it supported. The code isn’t really duplicated either. I factored out the main functionality of each function and the sync and a sync functions just call them in different ways

hipstersmoothie:
@edi9999 Switched to an implementation that just calls the current functions and return a promise. Similar to `Jimp.read`. I believe we should provide this functionality and not make the user do it themselves. Since we return promises in a few other places, it makes sense to do it in the last few places we can.

hipstersmoothie:
@edi9999 have i fixed your duplication concerns?

edi9999:
You can even just do it like this : 

```
const promisify = (fun, ctx, args) => {
  return new Promise((resolve, reject) => {
    args.push((err, data) => {
      err && reject(err);
      resolve(data);
    });
    fun.apply(ctx, args);
  });
};

 writeAsync(path) {
        return promisify(this.write, this, [path]);
 }
```

This way the "promisification code" is just in one place.

hipstersmoothie:
@edi9999 added `promisify` util function

Favna:
I see this closes #445 and some talk about `Jimp.create` but I can't find [any documentation](https://github.com/oliver-moran/jimp#creating-new-images) on this new function nor (easily) find it in the code changes by the PR. Could you clarify the usage?

hipstersmoothie:
You can just use `Jimp.read` to make a Jimp instance. It takes all the same args as the constructor. `Jimp.create` also exists but it just calls `Jimp.read`. I'll document it now

septs:
Request update "jimp.d.ts" file defines.

hipstersmoothie:
@septs what updates are you requesting ?

PR #492, opened 2018-07-29
Add module build

hipstersmoothie:
Module build (still has import/export in output) for ESM and webpack. The future is now!

Also in this PR:
* actually git ignore the build files. there was a space at the end that made the gitignore still include the files
* upgrade to babel@7
	* can use `.babelrc.js` to determine babel setup far more easily without repeating oneself
* remove `babel-preset-stage-0` and add plugins we actually use. the `stage-` presets are being deprecated
	* `@babel/plugin-proposal-class-properties`
	* `@babel/plugin-syntax-object-rest-spread`
* remove unneeded (default) config from karma.conf.js
* add `karma-browserify` to make the test run ridiculously fast. 
	* previously took a min or two to start up browser test
	* now it's almost instant
* remove unused stuff from `browser-build.js` (I want to delete most of this since we don't use much of it, another PR)
* only include `babel-plugin-source-map-support` for "development" env and not browser

Old time running `npm run test`: 43.04s
New time running `npm run test`: 72s

- [x] document module (es) build


hipstersmoothie:
@edi9999

.gitignore
```diff
@@ -41,5 +41,9 @@ node_modules
 
 package-lock.json
 
+# Browser Version
 browser/lib/*.js
```
edi9999: I think the files should be deleted if they are in .gitignore (.gitignore only applies to additions, not to modifications)

edi9999: The file browser/lib/jimp.min.js is still in the repository in your Pull Request : 

https://github.com/hipstersmoothie/jimp/blob/00093b99f32c1f71cc9be4aea613a0e0e52e7a45/browser/lib/jimp.min.js

Issue #493, opened 2018-07-29
Bundle Size is hUUUUUge


hipstersmoothie:
Our code: 143.84kb uncompressed, 16.14kb gzipped
All code: 1.43mb uncompressed, 158.21kb gzipped

😮 

https://bundlephobia.com/scan-results?packages=bignumber.js@7.2.1,bmp-js@0.1.0,buffer@5.0.2,exif-parser@0.1.9,file-type@8.1.0,jpeg-js@0.3.4,load-bmfont@1.2.3,mime@2.3.1,mkdirp@0.5.1,omggif@1.0.9,phin@2.9.0,pixelmatch@4.0.0,pngjs@3.0.0,tinycolor2@1.1.2,utif@2.0.0

```js
import Jimp from 'jimp/es'

Jimp.read('./exampleImages/npm.jpg').then(image => {
  image.resize(10, Jimp.AUTO).getBuffer(Jimp.AUTO, (err, res) => {
    console.log(res)
  })
})
```

<img width="1411" alt="screen shot 2018-07-28 at 10 54 38 pm" src="https://user-images.githubusercontent.com/1192452/43363421-4d8df60c-92b9-11e8-8bb8-27a1b7659a5e.png">

Offenders (only using stat sizes unless otherwise specified, cause parse time matters!):

* utif - TIFF support
	* total (including deps): 441.35kB 😭 119kB minified
	* size: 38kb
	* requires:
		* jpgjs = 190.02kB
		* pako = 213.33kB
	* potential replacements:
		* tiff@3.0.0 - 19.6kB minified
		* decode-tiff@0.2.1 - 3.8kB minified
		* both these have no encoder so we'd have to figure that out

* bignumber.js - used for hashes #497 
	* size:  84.19kB with tree-shaking 79.41kB, minified 17.7kB
	* potential replacements:
		* big.js - 5.9kB minified
		* buffer? maybe we don't even need the dep
		* @babel/plugin-syntax-bigint - this would be really cool. IDK how big it would be. the packaged minified is only 1.7kB though
			* https://developers.google.com/web/updates/2018/05/bigint

Image:

We use a bunch of different libraries to process images. Switching to a library that does multiple could provide a big win. Options

* images@3.0.1 - processes png, jpeg, and gif 9.8kB minified

* pngjs - encode/decode
	* potential replacements:
		* yapd@1.0.0 - 8.6kB minfified - low downloads though

Lower Hanging Fruit:

[...]

PR #496, opened 2018-07-30
Mime

hipstersmoothie:
The mime package supports a lot of mimes we don't care about. Coding our own save a bit from the total size of the library. Plus it's more explicit about what types are supported.

Our Version: 1.1kB transpiled unminified
Mime package: 9.9kb unminified

@edi9999 

related to #493


edi9999:
Thanks

PR #497, opened 2018-07-30
Switch away from BigNumber.js

hipstersmoothie:
BigNumber.js: size: 84.19kB with tree-shaking 79.41kB, minified 17.7kB
any-base:   : size: 2.68kB, 976B minified

I wrote tests to make sure I didn't break anything. Comparing the speed they are about the same.

related to #493


edi9999:
Very good job !

edi9999:
There are some conflicts now, but totally agree to merge this

PR #510, opened 2018-08-04
Errors

### LINKED ISSUE #347: Jimp.read never resolves or rejects
If you have an image with less than 8 Bytes, Jimp.read will never resolve or reject the promise. Leaving my application stuck.

I would expect Jimp to throw an "Invalid file signature" Error.

### LINKED ISSUE #353: Jimp stucking at a corrupted GIF file
I found this GIF file will cause Jimp stuck forever (right click to save):

Code snippet for reproducing this issue:
```js
const Jimp = require('jimp')

Jimp.read(`1.gif`, (err, image) => {
  console.log(image, err)
})
```

### LINKED ISSUE #408: Reading a non-image buffer prints weird error into console but returns a resolved promise
```
> require('jimp').read(new Buffer('foo')).then((x) => console.log('image', x))
mime must be a string
Promise { <pending> }
> image undefined
```

It should not print anything as a side effect. Instead the error should contained in the promise.

### LINKED ISSUE #420: error on jimp.read with special base64 image
After saving a base64 image as png, trying to read it with jimp.read causes the following error.

```
Error: There are some read requests waiting on finished stream
    at module.exports.ChunkStream._end (c:\Users\Baris\Desktop\github\nodebb_vscode\NodeBB\node_modules\pngjs\lib\chunkstream.js:100:7)
    at module.exports.ChunkStream.end (c:\Users\Baris\Desktop\github\nodebb_vscode\NodeBB\node_modules\pngjs\lib\chunkstream.js:86:10)
    at exports.PNG.PNG.end (c:\Users\Baris\Desktop\github\nodebb_vscode\NodeBB\node_modules\pngjs\lib\png.js:98:16)
    at exports.PNG.PNG.parse (c:\Users\Baris\Desktop\github\nodebb_vscode\NodeBB\node_modules\pngjs\lib\png.js:88:8)
    at Jimp.parseBitmap (c:\Users\Baris\Desktop\github\nodebb_vscode\NodeBB\node_modules\jimp\index.js:297:17)
    at c:\Users\Baris\Desktop\github\nodebb_vscode\NodeBB\node_modules\jimp\index.js:202:29
    at c:\Users\Baris\Desktop\github\nodebb_vscode\NodeBB\node_modules\graceful-fs\graceful-fs.js:78:16
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:511:3)
2018-03-22T12:38:44.435Z [9632] - error:  Error: invalid block type
    at Inflate.zlibOnError (zlib.js:153:15)
```

The code

```
jimp.read(pathToFile, function (err) {
    console.log(err);
});
```

Wrapping the jimp.read in a try catch doesn't catch the `Error:invalid block type` and the processes crashes.

Here is the image used.
94384f7f40dfd4b100d73fa8fe652c2e

### LINKED ISSUE #422: cannot trap corrupted jpg read errors
I am not able to handle situations where the jimp.read input image is a corrupted jpg. 
For example trying to read a 0 file length does not trigger any error.

```js
var Jimp = require("jimp");
var fs = require("fs");

try {
	Jimp.read("a.jpg", function (err, img) {
		try {
			if (err) 
				console.log("[Jimp.read] error loading ");
			else
				console.log("[Jimp.read] loaded ");
		} catch(e) {
			console.log("[Jimp.read] exception2:",e);
		}
	});
} catch(e) {
	console.log("[Jimp.read] exception:",e);
}
console.log("end");
```

### LINKED ISSUE #454: How to catch error 404
if i use Jimp.read("error url"), it will print error in console
 `Could not find MIME for Buffer <http://*/233> (HTTP: 404)`
i want catch it, then i can do another thing with it!

hipstersmoothie:
* better error catching in parseBitmap
* only emit error if there is no callback in constructor

closes #420 
closes #422 
closes #454 
closes #347 
closes #353 
closes #408 

Small code change, big impact 🎉

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -179,9 +179,9 @@ class Jimp extends EventEmitter {
 
             setTimeout(() => {
                 // run on next tick.
-                if (err) {
+                if (err && cb === noop) {
                     jimpInstance.emitError('constructor', err);
-                } else {
+                } else if (!err) {
                     jimpInstance.emitMulti('constructor', 'initialized');
                 }
 
--- a/src/modules/resize2.js
+++ b/src/modules/resize2.js
@@ -24,11 +24,9 @@ module.exports = {
     nearestNeighbor(src, dst) {
         const wSrc = src.width;
         const hSrc = src.height;
-        // console.log("wSrc="+wSrc + ", hSrc="+hSrc);
 
         const wDst = dst.width;
         const hDst = dst.height;
-        // console.log("wDst="+wDst + ", hDst="+hDst);
 
         const bufSrc = src.data;
         const bufDst = dst.data;
@@ -52,11 +50,9 @@ module.exports = {
     bilinearInterpolation(src, dst) {
         const wSrc = src.width;
         const hSrc = src.height;
-        // console.log("wSrc="+wSrc + ", hSrc="+hSrc);
 
         const wDst = dst.width;
         const hDst = dst.height;
-        // console.log("wDst="+wDst + ", hDst="+hDst);
 
         const bufSrc = src.data;
         const bufDst = dst.data;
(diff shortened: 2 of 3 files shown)


src/index.js
```diff
@@ -179,9 +179,9 @@ class Jimp extends EventEmitter {
 
             setTimeout(() => {
                 // run on next tick.
-                if (err) {
+                if (err && cb === noop) {
                     jimpInstance.emitError('constructor', err);
-                } else {
```
edi9999: I am not sure why you want to change the behavior of this ?

edi9999: (The rest looks good to me)

hipstersmoothie: If jimp emits the error then the callback will never be called with error, and the promise never catches the error. this change makes it so that jimp either 

1. calls the callback with the error
2. rejects with the error
3 or if no callback or promise is provided it emits the error

when we were calling emitError the whole execturion of the program would stop if you didn't have an `on('error')`. All the issues basically arose from just this.

PR #511, opened 2018-08-04
clean up getJimpDir

### LINKED ISSUE #281: getJimpDir() doesn't work as expected
When using jimp via npm, getJimpDir() returns './'.
https://github.com/oliver-moran/jimp/blob/master/index.js#L418
So it looks up the font directory in current directory.
But the font directory is in the node_modules/jimp directory.
And so it doesn't find the predefined fonts.

hipstersmoothie:
These regex were to work with karma. if block now handles this case so it can go

closes #281 - i think this was a problem with the regex


edi9999:
Good !

PR #191, opened 2016-11-11
Blendmodes

skoa-h:
I tried to implement it compatible. Had hard times with color burn & dodge and soft light, so they are still missing, but now we have quite some blend modes available.

PR #515, opened 2018-08-06
Blendmodes

### LINKED ISSUE #139: Blend mode argument for composite()
It would be nice to be able to blend images together instead of just straight pixel-copy compositing.

e.g. `image.composite( src, x, y, blendMode );`

`blendMode` could be `screen`, `multiply`, `overlay`, etc.

https://en.wikipedia.org/wiki/Blend_modes

hipstersmoothie:
# What's Changing and Why

Allows users to composite image in different ways similar to gimp.

closes #139
copied code from #191 

Nothing.

## Tasks

-   [x] Add tests
-   [x] Update Documentation
-   [x] Update `jimp.d.ts`


src/image-manipulation/effects.js
```diff
@@ -422,44 +430,70 @@ export function composite(src, x, y, cb) {
         return throwError.call(this, 'x and y must be numbers', cb);
     }
 
+    let { mode, opacitySource, opacityDest } = options;
```
edi9999: Can you add a test with these new params ?

hipstersmoothie: test added! it's only 1 can be more thorough if needed

edi9999: I would prefer to have a test that will expect the whole image to be equal to a base image so that we have more certainty that it will break fast if the code is wrong (like you've done in : https://github.com/oliver-moran/jimp/pull/524/files#diff-c632bfb61a4e129216f938dcb8d1a782R126)

hipstersmoothie: @edi9999 I added a better test but for some reason my bitmaps were always off by 1 before writing. The new test writes to the filesystem and checks those buffers.

PR #524, opened 2018-08-08
Unknown Characters

### LINKED ISSUE #523: text - writing special characters
For special characters that are not in the font, is it possible instead of Jimp ignoring them, to write a certain character (ex ツ would output as ?)

hipstersmoothie:
# What's Changing and Why

Jimp will write nothing to an image if an unknown char is found. Now it will render a '?'

closes #523

Printing

## Tasks

-   [x] Add tests
-   [x] Update Documentation
-   [x] Update `jimp.d.ts`


Alec-Fox:
for those looking for an easy fix for unknown characters (can can edit which characters dont get replaced in regex):
string = string.replace(/[^a-zA-Z0-9 _-]/g, '[?]');

hipstersmoothie:
or wait until the next release 😉

stephengardner:
Sorry to revive this but how can we allow Jimp to write non-latin characters, such as Chinese or Hebrew characters?

PR #530, opened 2018-08-09
BMP Color issue

### LINKED ISSUE #432: Jimp read bmp to base64 results in wrong colours
Hi,

When reading a BMP file using the .read method and then converting the image to a base64 string using the getBase64 method the image is not the same as the original image anymore. The weird thing is that it does not happen to all BMP's i'm converting, only a few.

Original BMP: 
screen shot 2018-04-18 at 13 38 55

Converted BMP:
screen shot 2018-04-18 at 21 38 06

Code:
```
const image = await Jimp.read("path/to/image.bmp");
image.getBase64(Jimp.MIME_BMP, (err, data) => { // Assigning data to image element });
```

Version: browser v0.2.27

Thanks in advance,
Smaïl

### LINKED ISSUE #521: Exporting PNG to BMP makes everything red
I'm trying to load a PNG, tweak it, and export it to a BMP.

When I do so, everything ends up very red.

My guess is that Jimp is passing an RGBA buffer to js-bmp, which is interpreting it as ABGR, which ends up throwing out the red, swapping blue and green, and using the alpha (0xFF) for all red pixels.

screen shot 2018-08-06 at 5 52 55 pm

hipstersmoothie:
# What's Changing and Why

`bmp-js` outputs the buffer in AGBR. This PR will convert bmp bitmap data to RGBA on encode and back to AGBR on encode. https://github.com/shaozilee/bmp-js

closes #521

pretty sure it also closes #432 

bmp file handling.

## Tasks

-   [x] Add tests
-   [x] Update Documentation
-   [x] Update `jimp.d.ts`


shawntabrizi:
@hipstersmoothie I believe I am running into this very issue right now using JIMP `v0.9.3`.

I read from a BMP, and make repeating writes to the file.

The writes are actually setting a batch of pixel values using the `bitmap.setPixelColour(color, x, y);` call.

What I notice is that every other write operation, the colors turn "red" like described in the issue. Then, the next write it turns back to normal.

I am wondering if the code somehow converts from AGBR -> RGBA and then back on a repating loop.

PR #534, opened 2018-08-09
Always return this for chaining

### LINKED ISSUE #527: Fix chaining
we take an optional callback but return it. this make it impossible to chain off of the this value. I propose changing all of these:

```js
    if (isNodePattern(cb)) {
        return cb.call(this, null, this);
    }
```

To this:

```js
    if (isNodePattern(cb)) {
       cb.call(this, null, this);
    }
```

which will allow the callback to be called but still allow chaining.

@edi9999 What do you think?

hipstersmoothie:
# What's Changing and Why

The `cb` returned nothing so there was no point in returning it. Instead return this for more chaining

closes #527 

Types! @VojtechStep did I do it right?

## Tasks

-   [x] Add tests
-   [x] Update Documentation
-   [x] Update `jimp.d.ts`

### DIFF
--- a/jimp.d.ts
+++ b/jimp.d.ts
@@ -234,284 +234,212 @@ declare namespace Jimp {
         toString(): string;
         getMIME(): string;
         getExtension(): string;
-        write(path: string, cb?: Jimp.ImageCallback): this;
+        write(path: string, cb: Jimp.ImageCallback): this;
         writeAsync(path: string): Promise<Jimp>;
-        deflateLevel(l: number): this;
-        deflateLevel<T>(l: number, cb: Jimp.ImageCallback<T>): T;
-        deflateStrategy(s: number): this;
-        deflateStrategy<T>(s: number, cb: Jimp.ImageCallback<T>): T;
-        filterType(f: number): this;
-        filterType<T>(f: number, cb: Jimp.ImageCallback<T>): T;
-        rgba(bool: boolean): this;
-        rgba<T>(bool: boolean, cb: Jimp.ImageCallback<T>): T;
-        quality(n: number): this;
-        quality<T>(n: number, cb: Jimp.ImageCallback<T>): T;
-        getBase64(mime: string, cb?: GenericCallback<string, any, this>): this;
+        deflateLevel(l: number, cb?: Jimp.ImageCallback<T>): this;
+        deflateStrategy(s: number, cb?: Jimp.ImageCallback<T>): this;
+        filterType(f: number, cb?: Jimp.ImageCallback<T>): this;
+        rgba(bool: boolean, cb?: Jimp.ImageCallback<T>): this;
+        quality(n: number, cb?: Jimp.ImageCallback<T>): this;
+        getBase64(mime: string, cb: GenericCallback<string, any, this>): this;
         getBase64Async(mime: string): Promise<Jimp>;
-        hash(base?: number): string;
-        hash<T>(cb: GenericCallback<string, T, this>): T;
-        hash<T>(
+        hash(cb?: GenericCallback<string, T, this>): this;
+        hash(
             base: number | null | undefined,
-            cb: GenericCallback<string, T, this>
-        ): T;
+            cb?: GenericCallback<string, T, this>
+        ): this;
         getBuffer(mime: string, cb: GenericCallback<Buffer>): this;
         getBufferAsync(mime: string): Promise<Buffer>;
-        getPixelIndex<T>(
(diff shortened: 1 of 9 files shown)


jimp.d.ts
```diff
@@ -234,369 +234,266 @@ declare namespace Jimp {
         toString(): string;
         getMIME(): string;
         getExtension(): string;
-        write(path: string, cb?: Jimp.ImageCallback): this;
+        write(path: string, cb: Jimp.ImageCallback): this;
```
VojtechStep: The callback is still optional

jimp.d.ts
```diff
@@ -234,369 +234,266 @@ declare namespace Jimp {
...
-        deflateStrategy(s: number): this;
-        deflateStrategy<T>(s: number, cb: Jimp.ImageCallback<T>): T;
-        filterType(f: number): this;
-        filterType<T>(f: number, cb: Jimp.ImageCallback<T>): T;
-        rgba(bool: boolean): this;
-        rgba<T>(bool: boolean, cb: Jimp.ImageCallback<T>): T;
-        quality(n: number): this;
-        quality<T>(n: number, cb: Jimp.ImageCallback<T>): T;
-        getBase64(mime: string, cb?: GenericCallback<string, any, this>): this;
+        deflateLevel(l: number, cb?: Jimp.ImageCallback<T>): this;
```
VojtechStep: You have to also remove the `<T>` part in the ImageCallback, since the type parameter is not defined anywhere

jimp.d.ts
```diff
@@ -234,369 +234,266 @@ declare namespace Jimp {
...
         getBufferAsync(mime: string): Promise<Buffer>;
-        getPixelIndex<T>(
+        getPixelIndex(
             x: number,
             y: number,
-            cb: GenericCallback<number, T, this>
-        ): T;
-        getPixelIndex(x: number, y: number, edgeHandling?: string): number;
-        getPixelIndex<T>(
+            cb?: GenericCallback<number, T, this>
```
VojtechStep: Same thing with the `T` parameter here and after, just replace it with `any`

jimp.d.ts
```diff
@@ -234,369 +234,266 @@ declare namespace Jimp {
...
+        resize(w: number, h: number, cb?: Jimp.ImageCallback<T>): this;
+        resize(
             w: number,
             h: number,
             mode: string | undefined | null,
-            cb: Jimp.ImageCallback<T>
-        ): T;
-        cover<T>(w: number, h: number, cb: Jimp.ImageCallback<T>): T;
-        cover<T>(
+            cb?: Jimp.ImageCallback<T>
```
VojtechStep: Since `cb` is optional here, you can make `mode` optional instead of `string | undefined | null`. The same goes for all the other ` | undefined | null` that are now followed by an optional callback

jimp.d.ts
```diff
@@ -234,369 +234,266 @@ declare namespace Jimp {
...
-        contain<T>(w: number, h: number, cb: Jimp.ImageCallback<T>): T;
-        contain(w: number, h: number, mode: string): this;
-        contain<T>(
+            cb?: Jimp.ImageCallback<T>
+        ): this;
+        contain(w: number, h: number, cb?: Jimp.ImageCallback<T>): this;
+        contain(
             w: number,
             h: number,
             mode: string,
```
VojtechStep: `mode` should be optional

jimp.d.ts
```diff
@@ -234,369 +234,266 @@ declare namespace Jimp {
...
-            cb: Jimp.ImageCallback<T>
-        ): T;
-        contain(w: number, h: number, alignBits: number): this;
-        contain<T>(
+            cb?: Jimp.ImageCallback<T>
+        ): this;
+        contain(
             w: number,
             h: number,
             alignBits: number,
```
VojtechStep: `alignBits` should be optional

jimp.d.ts
```diff
@@ -234,369 +234,266 @@ declare namespace Jimp {
...
-            cb: Jimp.ImageCallback<T>
-        ): T;
-        contain(w: number, h: number, alignBits: number, mode: string): this;
-        contain<T>(
+            cb?: Jimp.ImageCallback<T>
+        ): this;
+        contain(
             w: number,
             h: number,
             alignBits: number,
```
VojtechStep: Both `alignBits` and `mode` should be optional (it's valid to call `contain(20, 20, undefined, undefined, (err, img) => {})`)

PR #535, opened 2018-08-09
fix types

hipstersmoothie:
# What's Changing and Why

code review from @VojtechStep 

just types

## Tasks

-   [x] Add tests
-   [x] Update Documentation
-   [x] Update `jimp.d.ts`

### DIFF
--- a/jimp.d.ts
+++ b/jimp.d.ts
@@ -359,7 +359,7 @@ declare namespace Jimp {
         rotate(deg: number, cb?: Jimp.ImageCallback): this;
         rotate(
             deg: number,
-            mode: number | boolean,
+            mode: string | boolean,
             cb?: Jimp.ImageCallback
         ): this;
         flip(
@@ -479,7 +479,7 @@ declare namespace Jimp {
             src: Jimp,
             x: number,
             y: number,
-            options: Jimp.BlendMode,
+            options?: Jimp.BlendMode,
             cb?: Jimp.ImageCallback
         ): this;
         blit(src: Jimp, x: number, y: number, cb?: Jimp.ImageCallback): this;
--- a/jimp.d.ts
+++ b/jimp.d.ts
@@ -234,13 +234,13 @@ declare namespace Jimp {
         toString(): string;
         getMIME(): string;
         getExtension(): string;
-        write(path: string, cb: Jimp.ImageCallback): this;
+        write(path: string, cb?: Jimp.ImageCallback): this;
         writeAsync(path: string): Promise<Jimp>;
-        deflateLevel(l: number, cb?: Jimp.ImageCallback<T>): this;
-        deflateStrategy(s: number, cb?: Jimp.ImageCallback<T>): this;
-        filterType(f: number, cb?: Jimp.ImageCallback<T>): this;
-        rgba(bool: boolean, cb?: Jimp.ImageCallback<T>): this;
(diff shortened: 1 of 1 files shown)


hipstersmoothie:
@VojtechStep any idea how to fix the buffer thing?

jimp.d.ts
```diff
@@ -519,8 +519,8 @@ declare namespace Jimp {
             g: number,
             b: number,
             a: number,
-            cb: GenericCallback<number, T, Jimp>
-        ): this;
+            cb: GenericCallback<number, any, Jimp>
+        ): number;
```
VojtechStep: This should actually return `void`

jimp.d.ts
```diff
@@ -234,211 +234,211 @@ declare namespace Jimp {
...
+        colour(actions: ColorAction[], cb?: Jimp.ImageCallback): this;
 
         // Shape methods
-        rotate(deg: number, cb?: Jimp.ImageCallback<T>): this;
+        rotate(deg: number, cb?: Jimp.ImageCallback): this;
         rotate(
             deg: number,
-            mode: number | boolean | undefined | null,
-            cb?: Jimp.ImageCallback<T>
+            mode: number | boolean,
```
VojtechStep: It seems I've made a mistake before, the type of mode should be `string | boolean` and it should be optional

jimp.d.ts
```diff
@@ -447,42 +447,42 @@ declare namespace Jimp {
...
+        normalize(cb?: Jimp.ImageCallback): this;
+        invert(cb?: Jimp.ImageCallback): this;
+        gaussian(r: number, cb?: Jimp.ImageCallback): this;
         composite(
             src: Jimp,
             x: number,
             y: number,
-            options: Jimp.BlendModel,
-            cb?: Jimp.ImageCallback<T>
+            options: Jimp.BlendMode,
```
VojtechStep: Since we provide a default value for `options`, it should be marked as optional

jimp.d.ts
hipstersmoothie: this one for sure returns a number though

VojtechStep: Well [the function](https://github.com/oliver-moran/jimp/blob/master/src/index.js#L870) doesn't have a return statement

hipstersmoothie: I think you're reading the diff wrong. line 523 is a part of `rgbaToInt` not `appendConstructorOption`

VojtechStep: Oh 100% you're right, sorry about that 🤦‍♂️

VojtechStep:
@hipstersmoothie Regarding the Buffer thing, I'm not sure how to approach it. We should add `@types/node` as a dependency of some sort, but there are drawbacks:
- Regular dependencies: won't go here, types are only for development
- DevDependencies: If we put it here, they are installed even for users of the library who don't use TypeScript
- OptionalDependencies: They still get installed
- PeerDependencies: Users don't need to install the types, but the will get annoying messages in the terminal if they don't

hipstersmoothie:
Hmm. Better left to a different PR then. May just have to put something in the readme on how to configure it properly

nrkn:
You forgot to make this callback optional:

```
        static rgbaToInt(
            r: number,
            g: number,
            b: number,
            a: number,
            cb: GenericCallback<number, any, Jimp>
        ): number;
```

hipstersmoothie:
@nrkn please make a pr.

PR #542, opened 2018-08-10
fix cover function without alignBits

koprda:
# What's Changing and Why
fix cover function without alignBits
- Error: only use one flag per alignment direction

## Tasks

-   [ ] Add tests
-   [ ] Update Documentation
-   [ ] Update `jimp.d.ts`
-   [ ] Add [SemVer](https://semver.org/) Label


hipstersmoothie:
Thanks! Good catch

PR #547, opened 2018-08-13
Monorepo

hipstersmoothie:
# What's Changing and Why

Breaks jimp up into a set of many packages. Great news is this doesn't break the API at all!

Types - plugins that enable encoding/decoding for a format:
* `@jimp/type-png`
* `@jimp/type-jpeg`
* `@jimp/type-bmp`
* `@jimp/type-tiff`
* `@jimp/type-gif`

Motiviation: 
* Allows users to use other libraries for decoding/encoding easily. 
* Maybe someone doesn't care about JS only and wants to use wrappers around lower level image libraries.

Plugins - plugins that enable encoding/decoding for a format:
* `@jimp/plugin-resize`
* `@jimp/plugin-rotate`
* `@jimp/plugin-crop`
* `@jimp/plugin-blit`

Motiviation: 
* Allows users to write their own jimp function 
* Allows users to include and exclude whatever manipulation methods they want to

Other Packages
* `@jimp/core` - all the code you know and love from current jimp
* `@jimp/custom` - an API to customize a jimp configuration
* `jimp` - a customized jimp with all the current types and functions

# Todo

Make all functions into packages that are also plugins. This will enable people to write plugins for jimp (like `circle` or any other custom bitmap manipulation). It will also enable configuring a jimp the only does certain things.

Everything has been moved around so everthing

@edi9999 Input?

## Tasks

-   [x] Add tests
-   [ ] Update Documentation
-   [x] Update `jimp.d.ts`


hipstersmoothie:
@edi9999 I'd like some feedback on this. Do you think that its a good idea to structure the repo like this? The main benefits we get from this. 

* anyone can easily add functionality via a plugin
* you can use any encoder/decoder to make jimp perform better (for example using a C image decoder)

edi9999:
Good initiative. Jimp is indeed quite big even when you use only some of its modules. 

My personal use case of Jimp is just to have a simple to install decoder of multiple image types, and I use myself the array buffer that I pass to a qrcode decoder.

Maybe it would make sense to separate encoders from decoders, since depending on the usecase you would need only one or the others.

hipstersmoothie:
@edi9999 My thinking was that it made sense for the encoder and decoders to go together. Exposing the two functions together is quite easy and it doesn't add much if you aren't using the encoder (since the dep is already included). I could split them up but I think it adds a little to much overhead. It also makes sense for an image type to be a full plugin,

hipstersmoothie:
@edi9999 I've done everything now. I can probably add to each packages README a little more but I'm ready to merge when you are. the type def could also be a little better. but i am not an expert on that and it can be fixed going forward

✅ lint passing
✅ tests passing

edi9999:
Ok, indeed it makes no sense to split encoder/decoder if the underlying libraries do not split them. 

Let me know when you've finished the new packages you've talked about

hipstersmoothie:
@edi9999 all packages finished. once this is merged I'll make issues to document things a little more and get to work

hipstersmoothie:
I’m going to merge in the next 24 hours.  The tiff library is giving us a bunch of problems and this will help users be able to exclude that particular file type from their jimp

hipstersmoothie:
@edi9999

edi9999:
Sorry for the late reply, I'm  ok to merge too !

PR #572, opened 2018-08-27
Print

### LINKED ISSUE #506: Expose y position in print callback
I have to print 2 text blocks onto an image and the second one needs to be separated by exactly 100px after the first.

What I am doing right now is simply looping through the text and by using `measureText` function trying to find the number of lines that my text block has, then finding the max y value of each line of the text (which turns out to be 2px higher for every line than `font.common.lineHeight`, so I suppose there's some padding or something? Is it consistent?)
It's all kind of nice logically, BUT the way the text is being drawn on an image in the library itself is exactly by looping through all the text, counting the number of lines etc. So, I am doing practically same loop twice and increase the running time of my function twice :/

Is there a good way around this?

hipstersmoothie:
# What's Changing and Why

Expose final y value in print callback so users can easily stagger text

closes #506

nothing

## Tasks

-   [x] Add tests
-   [x] Update Documentation
-   [x] Update `jimp.d.ts`


edi9999:
If we expose the `y` value, we might as well expose the `x` value to have the full coordinate of the last letter, no ?

edi9999:
If you think the same, I would prefer to expose an object with both values : 

coordinate = {x: x, y: y}

hipstersmoothie:
@edi9999 comments addressed. Added the x value as well

PR #585, opened 2018-09-01
Has alpha

### LINKED ISSUE #271: image.hasAlpha() Method?
Is it possible to add a method for determining if a PNG has a transparency colour channel?  

Creating alpha channels when none is required results in unnecessarily large images.  

Ideally, the following could be used:
```js
image.rgba( image.hasAlpha() );
```
This would either preserve the channel when used remove it when unecessary.  

Alternatively you could use it to determine file types:
```js
if ( image.hasAlpha() ) {
    file = "filename.png";
} else {
    file = "filename.jpg";
}
```

hipstersmoothie:
# What's Changing and Why

determine if an image is using the alpha channel.

closes  #271

## Tasks

- [x] Add tests
- [x] Update Documentation
- [x] Update `jimp.d.ts`

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ An image processing library for Node written entirely in JavaScript, with zero n
 
 Installation: `npm install --save jimp`
 
+API documnetation can be found in the main [jimp package](./packages/jimp)
+
 ## Supported Image Types
 
 - [bmp](./packages/type-bmp)
@@ -37,6 +39,13 @@ Installation: `npm install --save jimp`
 - [resize](./packages/plugin-resize)
 - [rotate](./packages/plugin-rotate)
 - [scale](./packages/plugin-scale)
+
+## Custom Jimp
+
+If you want to extend jimp or omit types or functions visit [@jimp/custom](./packages).
+
+- Add file-types or switch encoder/decoders
+- Add add/remove plugins (image manipulation methods)
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ An image processing library for Node written entirely in JavaScript, with zero n
 
 Installation: `npm install --save jimp`
 
-API documnetation can be found in the main [jimp package](./packages/jimp)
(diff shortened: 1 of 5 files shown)


packages/jimp/test/filetypes.test.js
```diff
@@ -135,3 +135,17 @@ describe('FileType', () => {
     image.bitmap.data.should.be.deepEqual(expectedImg.bitmap.data);
   });
 });
+
+describe('hasAlpah', () => {
```
edi9999: alpha (typo)

packages/core/src/index.js
```diff
@@ -735,6 +735,27 @@ class Jimp extends EventEmitter {
...
+  hasAlpha() {
+    let hasAlpha = false;
+
+    for (let yIndex = 0; yIndex < this.bitmap.height && !hasAlpha; yIndex++) {
+      for (let xIndex = 0; xIndex < this.bitmap.width && !hasAlpha; xIndex++) {
+        const idx = (this.bitmap.width * yIndex + xIndex) << 2;
+        const alpha = this.bitmap.data[idx + 3];
+
+        if (alpha !== 0xff) {
+          hasAlpha = true;
```
edi9999: I'm not sure how well v8 and others read the code ahead, but wouldn't it be faster to return early ?

README.md
```diff
@@ -11,6 +11,8 @@ An image processing library for Node written entirely in JavaScript, with zero n
 
 Installation: `npm install --save jimp`
 
+API documnetation can be found in the main [jimp package](./packages/jimp)
```
edi9999: typo documentation

packages/core/src/index.js
hipstersmoothie: the for loops check for hasAlpha to be false so once if find an alpha channel both the loops break. I will go through every pixel if it doesn't find an alpha channel though.

hipstersmoothie:
@edi9999 comments addressed

PR #586, opened 2018-09-01
Distance from hash

### LINKED ISSUE #299: Support to compare image against a database of hashes?
I need to compare an image to a list of pre.calculated images hashes (`Jimp.hash()`).
I would like to use both `Jimp.distance()` and `Jimp.diff()`.
I.e: Is there any support to match an image against a (potentially BIG) list of saved images, without having to read every image in the list (and recalculate it's hash) again?

hipstersmoothie:
# What's Changing and Why

Calculate `distanceFromHash`

closes #299

nothing

## Tasks

- [x] Add tests
- [x] Update Documentation
- [x] Update `jimp.d.ts`

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -594,6 +594,18 @@ class Jimp extends EventEmitter {
   }
 
   /**
+   * Calculates the hamming distance of the current image and a hash based on their perceptual hash
+   * @param {hash} compareHash hash to compare to
+   * @returns {number} a number ranging from 0 to 1, 0 means they are believed to be identical
+   */
+  distanceFromHash(compareHash) {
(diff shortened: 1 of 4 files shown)


edi9999:
I think this API could be made even more reusable by allowing to pass two arguments : 

something like this : 

```
const hash1 = image1.pHash();
const hash2 = image2.pHash();
compareHashes(hash1, hash2);
```

This way, if for example you have 100 images and you want to compare them "all together" to remove duplicate or similar images, you have to compute the hashes only once.

Does this make sense ?

I don't know exactly how we can expose this.

hipstersmoothie:
@edi9999 that makes sense. I have added the `compareHashes` function/

PR #588, opened 2018-09-01
Print number

### LINKED ISSUE #568: Error: text must be a string or an object
Can you allow also numbers? Maybe just converted to string behind the scene?)

hipstersmoothie:
# What's Changing and Why

Allow print to print any type.

closes #568

## Tasks

- [x] Add tests
- [x] Update Documentation
- [x] Update `jimp.d.ts`


packages/plugin-print/src/index.js
```diff
@@ -263,13 +259,14 @@ export default () => ({
       let alignmentX;
       let alignmentY;
 
-      if (typeof text === 'object') {
+      if (typeof text === 'object' && text.text) {
```
edi9999: I think it should be : 

```
      if (typeof text === 'object' && typeof text.text === "string") {
```

since we would want : {text:''} , to not error.

edi9999: or maybe text.text != null

hipstersmoothie:
@edi9999 I changed it it !== null so a user could still provide non string in the text object

PR #598, opened 2018-09-03
Document re - configure jimp

hipstersmoothie:
# What's Changing and Why

Document that jimp custom can take an already configured jimp and extend it further

## Tasks

- [ ] Add tests
- [x] Update Documentation
- [ ] Update `jimp.d.ts`


edi9999:
Good !

PR #606, opened 2018-09-09
@jimp/cli

hipstersmoothie:
# What's Changing and Why

Makes jimp into a CLI where you can apply any image manipulation functions and even load custom plugins and types

todo:

- [x] load images for blit, composite, mask 
- [x] set up bin
- [x] create 

closes #606 

nothing

## Tasks

- [x] Add tests
- [x] Update Documentation
- [x] Update `jimp.d.ts`


hipstersmoothie:
@edi9999 what do you think about this? anything I should add or change?

PR #613, opened 2018-09-15
Blit src params Issue

### LINKED ISSUE #612: Regression in blit function when srcx is non-zero
# Expected Behavior

The blit function takes optional arguments `srcx, srcy, srcw, srch` to specify a region of the source image to blit. For example, if the source image is a 100w x 100h square and the destination is a 50w x 100h rectangle, it should be possible to blit the left half of the source image into the destination with `dest.blit(src, 0, 0, 0, 0, 50, 100)`, or the right half with `dest.blit(src, 0, 0, 50, 0, 50, 100)`.

This worked as expected in Jimp 0.2.28, but it no longer works in Jimp 0.4.0 (I haven't tested any other versions).

# Current Behavior

Blitting the left half of the source image (with `srcx = 0`) works as expected, but blitting the right half (with `srcx > 0`) doesn't affect the desination image at all. The problem seems to involve the following check at line 68 of `plugin-blit/dist/index.js`:

`if (x + sx >= 0 && y + sy >= 0 && maxw - x - sx > 0 && maxh - y - sy > 0)`

This check causes all pixels to be skipped when blitting the right half of the source image. I'm not sure whether the problem is the check itself, or the values of `sx` and `sy` that are being passed to the `scanQuiet()` callback.

## Context

- Jimp Version: 0.4.0 (bug does not exist in 0.2.28)
- Operating System: Linux
- Node version: 10.7.0

Thank you for this fantastic project!

hipstersmoothie:
# What's Changing and Why

respect src image offset

closes #612 

## Tasks

- [x] Add tests
- [x] Update Documentation
- [x] Update `jimp.d.ts`


hipstersmoothie:
@edi9999 After this is merged I'm going to do a release

PR #644, opened 2018-10-12
Correct mistype of variable name

tombull:
Mistyped variable name 'x' for 'w' meant that pixelated areas were always exactly as wide as how far across the image they started. Ooops.

# What's Changing and Why

I've changed the variable name to what it should be. So it works.

Nothing.

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`

### DIFF
--- a/packages/plugin-color/src/index.js
+++ b/packages/plugin-color/src/index.js
@@ -539,7 +539,7 @@ export default () => ({
 
     x = x || 0;
     y = y || 0;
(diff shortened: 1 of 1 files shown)


tombull:
Sorry for the sparse comments, it's a very simple and obvious change.

hipstersmoothie:
fixed in v0.5.5

PR #666, opened 2018-11-27
feat(gaussian): extract calculations from loops

ShynRou:
# What's Changing and Why

The Gaussian Blur Implementation currently has terrible performance, so I went a head and improved It.
I extracted all possible calculations out of the loops and added a weight map. 

Performance changes for me:
before 400x400 (5px radius) => 1600ms  
after    400x400 (5px radius) => 300ms

before 400x400 (30px radius) => 50s  
after    400x400 (30px radius) => 8.5s

### DIFF
--- a/packages/plugin-gaussian/src/index.js
+++ b/packages/plugin-gaussian/src/index.js
@@ -18,6 +18,19 @@ export default () => ({
     }
 
     const rs = Math.ceil(r * 2.57); // significant radius
+    const range = rs * 2 + 1;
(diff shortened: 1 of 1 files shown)


hipstersmoothie:
This still accomplishes the same type of blur right? there is also the 'blur' function, it's still different from that right?

ShynRou:
@hipstersmoothie Yes it does. I just calculate the weights per neighboring pixel beforehand.

PR #670, opened 2018-11-29
Add more details to the Align Modes in the doc

joseph4tw:
# What's Changing and Why

I was having trouble recently figuring out how to center some text on an image and I kept thinking that the `Jimp.HORIZONTAL_ALIGN_CENTER` meant that the cursor position of the text object would be at its horizontal center, not that the position of the x-axis on the image to the center. After a bunch of trial and error, I finally realized that these align modes change the position of the axis that you're specifying.

I figured that adding this table and adjusting the example for `print` might help other users who may run into this issue.

Nothing, just docs :)

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`

### DIFF
--- a/packages/jimp/README.md
+++ b/packages/jimp/README.md
@@ -316,6 +316,17 @@ Jimp.VERTICAL_ALIGN_MIDDLE;
 Jimp.VERTICAL_ALIGN_BOTTOM;
 ```
 
+Where the align mode changes the position of the associated axis as described in the table below.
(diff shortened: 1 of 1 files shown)


hipstersmoothie:
Thanks for the change!

joseph4tw:
Sure thing, I'm glad I could help :)

PR #672, opened 2018-12-04
Beautify documentation of print plugin

f-barth:
# What's Changing and Why
An enumeration in README of plugin-print did use `_` instead if `-` which lead to weird rendering

nothing

## Tasks

- [x] Update Documentation


hipstersmoothie:
Thanks!

PR #715, opened 2019-03-29
Fix bug in autocrop with leaveBorder enabled

f-barth:
# What's Changing and Why
* Centralise application of border for a better overview 
* Add safety check for crops to be greater than or equal to 0
* Currently the crop size is always reduced by `leaveBorder` value, which can lead to negative values if the crop size is smaller than the border.

nothing

## Tasks

- [x] Add tests
- [x] Update Documentation => not needed
- [x] Update `jimp.d.ts` => not needed

### DIFF
--- a/packages/plugin-crop/test/autocrop.test.js
+++ b/packages/plugin-crop/test/autocrop.test.js
@@ -5,7 +5,7 @@ import crop from '../src';
 
 const jimp = configure({ plugins: [crop] }, Jimp);
 
-describe('Autocrop', () => {
+describe.only('Autocrop', () => {
   it('image with transparent surround color', async () => {
     const imgSrc = await jimp.read(
       mkJGD(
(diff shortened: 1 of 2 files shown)


packages/plugin-crop/test/autocrop.test.js
```diff
@@ -5,7 +5,7 @@ import crop from '../src';
 
 const jimp = configure({ plugins: [crop] }, Jimp);
 
-describe('Autocrop', () => {
+describe.only('Autocrop', () => {
```
hipstersmoothie: remove this only please

hipstersmoothie:
Looks good! Just remove the only and I'll merge + release

packages/plugin-crop/test/autocrop.test.js
f-barth: done, sorry for letting that slip :)

jagaSto:
@hipstersmoothie, any progress on this? :)

PR #741, opened 2019-05-21
Fix cropping full width slices

### LINKED ISSUE #675: crop() works incorrectly when x = 0 and w = current image width
# Expected Behavior

Running crop() should result in a buffer of the appropriate length

# Current Behavior

Running crop() results in a buffer with an incorrect length

There is a bug on this line:
https://github.com/oliver-moran/jimp/blob/master/packages/plugin-crop/src/index.js#L30

Specifically, it looks like we're shifting things left by 3 bits instead of 2 bits.  Assuming that we have 4-byte pixels (RGB+alpha), we should only shift by 2 bits.  (2**2 = 4)

## Steps to Reproduce

```
$ node
> const Jimp = require('jimp');
undefined
> t = new Jimp(100, 100, 0x00000000);
<Jimp 100x100>
> t.bitmap.data.length == 4*100*100
true
> t.clone().crop(0, 0, 100, 50).bitmap.data.length == 4*100*50
false
```

(That last line should report true!)

## Context

- Jimp Version: 0.6.0
- Operating System: Ubuntu Linux 18.04
- Node version: 8.14.0

## Fix

I'm not going to hassle with making a pull-request since this is a one-line fix.  Basically, change the bit shift to 2 instead of 2+1.  Specifically, the line should read:
`const end = (start + h * w) << 2;`

NiGhTTraX:
# What's Changing and Why

Fixes #675.

Changes are limited to the crop plugin.

## Tasks

- [x] Add tests: I've added a few tests that check cropping full width/height slices from various starting positions. It doesn't cover the entire crop plugin but I think it's a good start.
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`
- [ ] Add [SemVer](https://semver.org/) Label - should I bump the patch version?

### DIFF
--- a/packages/plugin-crop/src/index.js
+++ b/packages/plugin-crop/src/index.js
@@ -27,7 +27,7 @@ export default function pluginCrop(event) {
     if (x === 0 && w === this.bitmap.width) {
       // shortcut
       const start = (w * y + x) << 2;
-      const end = (start + h * w) << (2 + 1);
+      const end = (start + h * w) << 2;
 
       this.bitmap.data = this.bitmap.data.slice(start, end);
     } else {
--- a/packages/plugin-crop/test/crop.test.js
+++ b/packages/plugin-crop/test/crop.test.js
@@ -1,0 +1,110 @@
+import { Jimp, mkJGD } from '@jimp/test-utils';
+import configure from '@jimp/custom';
+
+import crop from '../src';
+
+const jimp = configure({ plugins: [crop] }, Jimp);
+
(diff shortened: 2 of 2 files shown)


hipstersmoothie:
@NiGhTTraX  please rebase so the CI runs

NiGhTTraX:
@hipstersmoothie done

hipstersmoothie:
:rocket: PR was released in v0.6.6 :rocket:

PR #746, opened 2019-06-04
Fix: #745 Tyepscript definition wrong return type on hash()

### LINKED ISSUE #745: Typescript definition - Wrong hash() return type
# Expected Behavior

According to the API document, we shall just use
``` typescript
let hashString: string = image.hash();
```
Or according to the `jimp.d.ts`
``` typescript
image.hash(hash => {
    hashString = hash;
};
```

# Current Behavior

`image.hash()` return `Jimp` object. And the callback function return a type `null | Error`

- Jimp Version: 0.6.4
- Operating System: macOS 10.14.5
- Node version: node v11.3.0

soimy:
# What's Changing and Why
According to #745, `jimp.d.ts` is having wrong return type `this` on `hash()`, which should be `string`
This PR fixed this.

Nothing

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [x] Update `jimp.d.ts`

closes #745

### DIFF
--- a/packages/jimp/jimp.d.ts
+++ b/packages/jimp/jimp.d.ts
@@ -126,11 +126,11 @@ interface Jimp {
   quality(n: number, cb?: ImageCallback): this;
   getBase64(mime: string, cb: GenericCallback<string, any, this>): this;
   getBase64Async(mime: string): Promise<string>;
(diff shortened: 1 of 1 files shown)


hipstersmoothie:
:rocket: PR was released in v0.6.5 :rocket:

PR #770, opened 2019-08-06
Made typings plugin friendly & add typings for every package

crutchcorn:
# What's Changing and Why
Previously, there was a single index.d.ts file in `jimp`. However, if you wanted to use any of the `@jimp` files, you were unable to get the typings for them individually.

This meant a few things:
- You couldn't get typing info if you were using `@jimp/custom`
- The definition files were often in different places from their logic, making maintaining them much harder

The second point was made even clearer as this went on as there were a few plugins that had no type information previously (thusly making the `jimp` package type info incomplete). As a result, the following plugins now have typings on the `jimp` package that were previously not present:

- Circle
- Fisheye
- Shadow
- Threshold

This will likely be a breaking change for `jimp` (unless we want to export the same typings as before using `export * from '@jimp/core` or something similar)

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [x] Update `jimp.d.ts`

Published PR with canary version: `0.7.1-canary.770.193.0`


hipstersmoothie:
@crutchcorn  please rebase so the CI runs

crutchcorn:
@hipstersmoothie done

crutchcorn:
Wanted to check on the status of this. Seems that tests have passed

hipstersmoothie:
@crutchcorn sorry! I'm the only maintainer on this and really don't use jimp at all in my work. So I'm a little slow at getting to the issues. I haven't had time to take a proper look at this, but I'll look at it tomorrow

crutchcorn:
No worrier in the slightest. I know how maintaining something like this that we don't use ourselves goes. Just wanted to give a small bump cuz I know I need reminders on the one I maintain 😁

hipstersmoothie:
Really liking what I'm seeing though.

I'd prefer to make the change as breakingless as possible. Would that be easy?

crutchcorn:
Actually, if I did this properly this is not a breaking change in terms of the main Jimp typings. If you'd like, I can export the same other typings (ala the font options, etc) and then it will be 100% non-breaking change

hipstersmoothie:
Yeah do that and I think this is good to merge

crutchcorn:
Actually, you know what? I was just about to make that "fix", but then realized that none of the other types in that file were actually exported, so this PR in it's current state is already not a breaking change.

Pocky9744:
how can i convert .jpg to .png?
i want to convert .jpg image before color picking.
because .jpg's color picking is not exact.

crutchcorn:
Mmm, apologies. I'm going to reject my own PR until I can fix some issues with it. I'll try to have them fixed in the next day or two. Huge apologies for bumping when things aren't fully working >~<

crutchcorn:
So sorry once again for bumping before being ready.

I have now double and tripple checked that this should work in my own project which uses the `configure` function. I have also checked that the original `jimp` package exports the same things in it's typings now. This PR should be GTG and does not introduce any breaking changes

crutchcorn:
Actually, while this can totally be a patch for `jimp`, I might suggest making it a minor bump as it technically introduces the feature of typings to all of the non `jimp` mainline package

crutchcorn:
Merge conflicts resolved, `scanIterator` added to the `@jimp/core` package (and as a result the `jimp` typings)

Still not a breaking change :)

hipstersmoothie:
:rocket: PR was released in v0.8.0 :rocket:

PR #781, opened 2019-08-31
Feature scanIterator

ozelot379:
# What's Changing and Why
Currently with `image.scan` it's possible to iterate over the image with a callback

But on some cases it's better if it would possible to iterate with a `for` loop, like:

```js
for (const { x, y, idx, image } of image.scanIterator(0, 0, image.bitmap.width, image.bitmap.height)) {
    ...
}
```

An advantage would be too, it's possible to abort the scan with `break;`, with the callback this is not possible (currently)

~~I also change `image.scan` to use internal the new `scanIterator` and call the callback~~

## Tasks

- [X] Add tests
- [x] Update Documentation
- [X] Update `jimp.d.ts`

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -2,7 +2,7 @@ import fs from 'fs';
 import Path from 'path';
 import EventEmitter from 'events';
 
-import { isNodePattern, throwError, scan } from '@jimp/utils';
+import { isNodePattern, throwError, scan, scanIterator } from '@jimp/utils';
 import anyBase from 'any-base';
 import mkdirp from 'mkdirp';
 import pixelMatch from 'pixelmatch';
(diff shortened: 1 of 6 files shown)


hipstersmoothie:
Does this have any performance implications? I'd be hesitant to make this the default if it makes things slower since `scan` is used by almost every function.

ozelot379:
@hipstersmoothie Thanks for your hint

I made some mess tests (10000px x 10000px Jimp image) and I take values like:

```
Mess scanCallback: 500ms
Mess scanCallbackWithIterator: 2933ms
Mess scanIterator: 2338ms
```

So you're right and the iterator is very slow

So I revert the `scan` function change

hipstersmoothie:
:rocket: PR was released in v0.7.0 :rocket:

PR #782, opened 2019-09-01
Replace `npm` usage with `yarn`

ozelot379:
# What's Changing and Why
Seems this project uses `yarn` for package management

So I replaced some `npm` usages with `yarn` in documentation to avoid 'side effects', because other parts of the file already calls `yarn` and not `npm`

Nothing

## Tasks

- [ ] Add tests
- [X] Update Documentation
- [ ] Update `jimp.d.ts`


hipstersmoothie:
:rocket: PR was released in v0.6.5 :rocket:

Issue #763, opened 2019-07-24
[Feature] Add optional divisor to convolute()


OutsourcedGuru:
**Is your feature request related to a problem? Please describe.**
When searching the Internet for convolution kernels, one often finds some which normalize at the end of the calculation with division.

A typical example for a Gaussian blur might look like this:

```
| 1   4   6   4   1 |
| 4  16  24  16   4 |
| 6  24  36  24   6 |
| 4  16  24  16   4 |
| 1   4   6   4   1 | / 256
```

To find the equivalent version which doesn't include a divisor step, one would need to laboriously divide each of those integers with the divisor, resulting in something which might look like this:

```
[0.00390625,   0.015625,   0.0234375,   0.015625,   0.00390625],
[0.015625,     0.0625,     0.09375,     0.0625,     0.015625],
[0.0234375,    0.09375,    0.140625,    0.09375,    0.0234375],
[0.015625,     0.0625,     0.09375,     0.0625,     0.015625],
[0.00390625,   0.015625,   0.0234375,   0.015625,   0.00390625]
```

**Describe the solution you'd like**
What would be nice would be if the `convolute()` functions added another (optional) argument called `divisor` which defaults to `1`.

**Describe alternatives you've considered**
Noting that my app is a Node.js daemon, I was able to monkey-patch `node_modules/@jimp/plugin-color/dist/index.js` inside the `convolute()` function definition like this:
```
      var ksize = (kernel.length - 1) / 2;
      x = isDef(x) ? x : ksize;
      y = isDef(y) ? y : ksize;
      w = isDef(w) ? w : this.bitmap.width - x;
      h = isDef(h) ? h : this.bitmap.height - y;
      var source = this.cloneQuiet();
      // Begin edit ------------------
      var divisor = 256; // Imagine instead that the function accepted it as an argument
      // End edit --------------------
      this.scanQuiet(x, y, w, h, function (xx, yx, idx) {
        var value = applyKernel(source, kernel, xx, yx);
        // Begin edit ------------------
        value[0] /= divisor;
        value[1] /= divisor;
        value[2] /= divisor;
        // End edit --------------------
        this.bitmap.data[idx] = this.constructor.limit255(value[0]);
        this.bitmap.data[idx + 1] = this.constructor.limit255(value[1]);
        this.bitmap.data[idx + 2] = this.constructor.limit255(value[2]);
      });
```

[...]

PR #790, opened 2019-09-10
Og convolute divisor

OutsourcedGuru:
# What's Changing and Why
This addresses FR #763 for convolution divisors. Apologies if it's not 100% how you'd have done it but I've tested two scenarios and it works as expected.

You could change the order of the arguments so that this is near the end. Honestly, I don't love these anonymous incoming variables which might be _this_ or _that_ depending upon the order. :P

Testing should be performed when the user includes the other optional arguments.

## Tasks

- [ ] Add tests (?)
- [x] Update Documentation
- [ ] Update `jimp.d.ts` (?)
- [ ] Add [SemVer](https://semver.org/) Label (?)

## Test code I used
```
# Within the context of a Node Express app's router in this case...
router.get('/', function(req, res, next) {
  var kernel =  [
                  [1,  4,  6,  4, 1],
                  [4, 16, 24, 16, 4],
                  [6, 24, 36, 24, 6],
                  [4, 16, 24, 16, 4],
                  [1,  4,  6,  4, 1]
                ];
  var divisor = 256;
  Jimp.read('test.png', function(errRead, fileIn) {
    if (errRead) throw errRead;

    // Also tested this by omitting the divisor argument below
    // which results in a very white output, as expected
    fileIn.convolute(kernel, divisor, function(errConvolute, fileAdjusted) {
      if (errConvolute) throw errConvolute;

      fileAdjusted.write('blurred.png', function(errWrite) {
        if (errWrite) throw errWrite;

        res.render('index', { title: 'Done' });
      });
    });
  });
});
```

### DIFF
--- a/packages/plugin-color/README.md
+++ b/packages/plugin-color/README.md
@@ -240,6 +240,7 @@ Applies a convolution kernel to the image or a region
 - @param {number} y (optional) the y position of the region to apply convolution to
 - @param {number} w (optional) the width of the region to apply convolution to
 - @param {number} h (optional) the height of the region to apply convolution to
+- @param {number} divisor (optional) the final divisor to be applied to each value
 - @param {function(Error, Jimp)} cb (optional) a callback for when complete
 
 ```js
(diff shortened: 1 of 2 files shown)


hipstersmoothie:
@OutsourcedGuru This is currently a breaking change since the new prop must come first. If you put divisor as the last argument I am pretty sure you can achieve a non breaking change. For this to get merged you must at least get the old tests passing

Issue #736, opened 2019-05-14
Issue with Image size 3024x4032


RobertCalamari:
# Expected Behavior

I am just trying to take an image from one folder and move it to another folder.

# Current Behavior

However, I am having an issue with images that are size 3024x4032. When my code is ran and I check for the width and the height, it looks as if they switched. Code is below.

## Steps to Reproduce

The code takes 5 images named "pics (_n_).jpg" and moves them to a new folder. 

function runchanger(num, callback){
    if(num == 0){
        console.log("DONE");
    }
    else{
        var filetoget = './client/img/imgprcss/tochange/pics (' + num + ').jpg';
         var filechanged = './client/img/imgprcss/changed/keeping (' + num + ').jpg';

         Jimp.read(filetoget, (err, image) => {
              if (err) throw err;
              console.log(num + " - were her changing the image - width:" + image.bitmap.width + " | height: " + image.bitmap.height);
                image.write(filechanged);
                callback(num-1, runchanger);
        });     
        
    }
    
}
     
runchanger(5, runchanger);

IF YOUR ISSUE DEPENDS ON A PARTICULAR IMAGE BE SURE TO INCLUDE THIS AS WELL. WE CAN'T REPORDUCE IF WE DON'T HAVE YOUR IMAGE

Images are attached. I included some that are fine to show that not all of them are effected. 
Here are the images before processing:
![pics (396)](https://user-images.githubusercontent.com/36252593/57721391-0da7b000-7652-11e9-9506-91629bcb1ee9.JPG)
![pics (394)](https://user-images.githubusercontent.com/36252593/57721329-e7821000-7651-11e9-960b-6c7540d5f116.JPG)
![pics (395)](https://user-images.githubusercontent.com/36252593/57721330-e7821000-7651-11e9-8510-67d581bb4238.JPG)
![pics (399)](https://user-images.githubusercontent.com/36252593/57721331-e81aa680-7651-11e9-8b04-68541cbe55d8.JPG)
![pics (400)](https://user-images.githubusercontent.com/36252593/57721332-e81aa680-7651-11e9-933f-cb9272edb86d.JPG)

Here are the images after:
![keeping (395)](https://user-images.githubusercontent.com/36252593/57721410-1c8e6280-7652-11e9-89ce-e8a8e0c0173f.jpg)
![keeping (396)](https://user-images.githubusercontent.com/36252593/57721411-1c8e6280-7652-11e9-8900-115781c213cb.jpg)
[...]

Issue #733, opened 2019-05-12
Automatic image orientation based on EXIF is broken


mkondel:
# Expected Behavior
When processing any image, the intended orientation should be preserved.

# Current Behavior
In the provided example, a portrait image gets saved in landscape. You can see the black bars that get added on left and right. The core image is oriented correctly inside, but the output file has the wrong dimensions.

# Failure Information (for bugs)

## Steps to Reproduce
1. git clone https://github.com/mkondel/jimp-exif-text.git
2. yarn
3. yarn start
4. look at input.jpg and output.jpg

**Screenshots**
I have screenshots in the repo linked above ^^.

## Context
I am using Jimp with excellent result on AWS Lambda. This bug is not related to that at all, as it can be replicated locally.

- Jimp Version: v0.6.4
- Operating System: Linux (Fedora)
- Node version: v8.12.0

m-jiang:
I am also experiencing this issue, which is quite crippling for any application that allows users to upload an image from their device and camera.

This issue has noting to do with the rotate function (yet). As in the sample code provided by @mkondel, simply loading and writing the image will already deform it.

mkondel:
I fixed my issue by using another lib called [jpeg-autorotate](https://github.com/johansatge/jpeg-autorotate). Please checkout my repo above for a working demo. I use this lib before JIMP to "burn in" the rotation.

```javascript
const fileIn = fs.readFileSync('input.jpg')
const jo = require('jpeg-autorotate')
const {buffer} = await jo.rotate(fileIn, {quality: 30})
const image = await Jimp.read(buffer)
```

DjulLau:
I agree with @grnch : it would be great if this could be fixed in Jimp.

I've struggled with this for hours yesterday wondering what I was doing wrong.
[jpeg-autorotate](https://github.com/johansatge/jpeg-autorotate) works but you have to add another module and understand how it works (for exemple it will throw an error (!) if no rotation has to be done so you need another code path to handle this case...).

sesam:
The fix is merged 🎉 and was released in  v0.8.5 🚀

So can this ticket be closed?

mixdesign:
@oliver-moran I think this is broken again in v0.14.0 (maybe before)

VolodymyrKornetskyi:
The same issue with v0.14.0

LeonardodeLima:
this issue with v0.16.1

dosomder:
Any news?

rwichmann:
[...]

PR #791, opened 2019-09-12
Image dimensions during exif rotation have been corrected

popinguy:
# What's Changing and Why
Fixes https://github.com/oliver-moran/jimp/issues/736 and https://github.com/oliver-moran/jimp/issues/733

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`

Published PR with canary version: `0.8.1-canary.791.466.0`

### DIFF
--- a/packages/core/src/utils/image-bitmap.js
+++ b/packages/core/src/utils/image-bitmap.js
@@ -27,7 +27,7 @@ function getMIMEFromBuffer(buffer, path) {
 /*
  * Automagically rotates an image based on its EXIF data (if present)
  * @param img a constants object
-*/
+ */
 function exifRotate(img) {
   const exif = img._exif;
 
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/image-bitmap.js
```diff
@@ -46,16 +46,16 @@ function exifRotate(img) {
         img.mirror(false, true);
         break;
       case 5: // Mirror horizontal and rotate 270 CW
-        img.rotate(-90, false).mirror(true, false);
```
grnch: The `false` flag should also be removed under `case 3:` above. I know it doesn't make a difference in case of upside down rotation (180 degrees), but it should still be removed for consistency.

(GitHub doesn't let me comment on that line directly because it's too far from the diff)

hipstersmoothie:
:rocket: PR was released in v0.8.5 :rocket:

PR #660, opened 2018-11-19
Improve TypeScript declarations file

Favna:
# What's Changing and Why

Addresses but does not fully close #581

Proper way to import is now `import * as X from 'jimp'` wherein X can be
whatever the end-user wants, in most cases `Jimp`.

As stated on the aforementioned issue another PR might come with a more thorough fix not requiring `// @ts-ignore`

## What else might be affected

Method of importing with TypeScript, possible requires doc updates.

## Tasks

- [ ] Add tests
- [x] Update Documentation
- [ ] Update `jimp.d.ts`
- [x] Add [SemVer](https://semver.org/) Label

Issue #581, opened 2018-08-29
error TS2497: Module ''jimp'' resolves to a non-module entity and cannot be imported using this construct.


AtaS:
Hi,

I cannot import jimp with typescript, getting error. Details are below.

# Expected Behavior
There shouldn't be a compilation-time error.

# Current Behavior

There is a compilation-time error.
`error TS2497: Module ''jimp'' resolves to a non-module entity and cannot be imported using this construct. import * as jimp from 'jimp';`

Also,  `import jimp from 'jimp';` doesn't work because it imports the jimp into jimp_1 and becomes not usable...

# Failure Information (for bugs)

## Steps to Reproduce
Import jimp in a typescript file with `import * as jimp from 'jimp';`

- Jimp Version: ^0.3.9
- Operating System: MacOS
- Node version: v8.11.3
- TypeScript version: 3.0.1

## Failure Logs

`error TS2497: Module ''jimp'' resolves to a non-module entity and cannot be imported using this construct. import * as jimp from 'jimp';`

hipstersmoothie:
released in v0.3.10

hipstersmoothie:
This actaully doesn't seem to work fully.

Now when i try to import in a non ts file it is exported under default

Do you know of any way to get this to work @AtaS ?

```
declare module 'jimp' {
  export = Jimp.Jimp;
  export default Jimp.Jimp;
}

```

hipstersmoothie:
I gotta go back to because the latest release broke vscode's autocompletion. 

```ts
declare module 'jimp' {
  export = Jimp.Jimp;
}
```

I've found you can make <0.3.10 work with the following 

```ts
import Jimp = require('jimp');
```

hipstersmoothie:
If I `export class Jimp {` and `export default Jimp.Jimp` i get autocompletions in CJS code and tsc compiles. but tyepscript has no autocompletions...

hipstersmoothie:
I've found if you do either of the following in your tsconfig.json <0.3.10 `tsc build` works and autocompletions in both environments work

```json
  "compilerOptions": {
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true
  },
```

So with that I am going to revert back to the code in 0.3.9

hipstersmoothie:
Feel free to make a or!! I’m pretty new to typescript

hipstersmoothie:
@AtaS ping

AtaS:
[...]

PR #792, opened 2019-09-12
Fix issues with typings using classes, publish @core typings, and fix 3.1 typings

crutchcorn:
# What's Changing and Why

Copied from https://github.com/oliver-moran/jimp/issues/785#issuecomment-530852049

> When doing testing for #786 , in order to avoid issues with yarn link I copied + pasted the packages folder. As a result, I forgot to update the npm whitelist and both @jimp/core and jimp do not copy their types folder to the final build, thus causing the error above. I was able to repro the issue and resolve it only by copying the types folder of each into their respective node_modules folder.

## Unrelated notes on TS that I noticed while debugging this PR

There were two more things that I wanted to bring up here (just to keep everyone up-to-date, neither of these issues are breaking or should be handled now):

1) Not introduced in 0.8 - `require` no longer works for typings. This may have been noticed by @hipstersmoothie even when trying to fix tests in 0.6: https://github.com/oliver-moran/jimp/commit/7dff2878be3fadc346d15d954bbc9fd7bd749ae7#diff-9687f7d238d1b6cbaf5bddfd10b57491L5

The reason `const Jimp = require('jimp')` doesn't work with types anymore is because now it's under `Jimp.default`. Now, if we wanted to enable this functionality again, we could easily change `index.d.ts` file from `jimp` like so:

```
export = Jimp;
```

But then the problem is that, as the TS language extensions will tell you, you cannot export like this and have any other exports. To get around this in the past, we used namespaces:
https://github.com/oliver-moran/jimp/blob/14055cf60e66701987bb148a2d389844dd0e9cb4/packages/jimp/jimp.d.ts

So if we're wanting to restore said functionality, we'd need to do the same thing today. Otherwise, if we're okay with it being as it is (the `import` works just fine), we should likely remove the `require` example from the README

2) There seems to have been an invisible requirement developed for these types: `@types/node`. Now, again, this was not introduced in 0.8 and it's actually somewhat hard to pinpoint when this was introduced (the type in question that it relies on is `Buffer`. Now, there's a few things we could do here:

- Introduce a peer dependency for the `@types/node` typings to inform the user that we expect them to have it installed
- Copy + paste the `Buffer` type ~ due to (seemingly) only having one type from that package we depend on we could likely just copy the type to be local. The issue here is that maintanance and type interop are a big problem - I'd highly suggest against this one
- Do nothing ~ I've seen a lot of typings simply assume that these types are installed to not enforce a peer dependency on non-TS users

Published PR with canary version: `0.8.2-canary.792.360.0`

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -9,7 +9,8 @@
     "dist",
     "es",
     "index.d.ts",
-    "fonts"
+    "fonts",
+    "types"
   ],
   "repository": {
     "type": "git",
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -2,8 +2,8 @@ export * from './etc';
 export * from './functions';
 export * from './plugins';
 export * from './utils';
-import {Jimp} from './jimp';
+import {Jimp, JimpConstructors} from './jimp';
 
-export { Jimp };
+export { Jimp, JimpConstructors };
 declare const defaultExp: Jimp;
 export default defaultExp;
--- a/packages/core/types/jimp.d.ts
+++ b/packages/core/types/jimp.d.ts
@@ -10,22 +10,24 @@ import {
   RGB
 } from './etc';
 
-export declare class Jimp {
-  // Constructors
-  constructor(path: string, cb?: ImageCallback);
-  constructor(urlOptions: URLOptions, cb?: ImageCallback);
-  constructor(image: Jimp, cb?: ImageCallback);
-  constructor(data: Buffer, cb?: ImageCallback);
-  constructor(data: Bitmap, cb?: ImageCallback);
-  constructor(w: number, h: number, cb?: ImageCallback);
-  constructor(
+export interface JimpConstructors {
+  new(path: string, cb?: ImageCallback): this;
+  new(urlOptions: URLOptions, cb?: ImageCallback): this;
+  new(image: Jimp, cb?: ImageCallback): this;
+  new(data: Buffer, cb?: ImageCallback): this;
+  new(data: Bitmap, cb?: ImageCallback): this;
+  new(w: number, h: number, cb?: ImageCallback): this;
+  new(
     w: number,
     h: number,
     background?: number | string,
     cb?: ImageCallback
-  );
+  ): this;
   // For custom constructors when using Jimp.appendConstructorOption
-  constructor(...args: any[]);
+  new(...args: any[]): this;
+}
+
+export interface Jimp extends JimpConstructors {
   prototype: this;
   // Constants
   AUTO: -1;
--- a/packages/core/types/jimp.d.ts
+++ b/packages/core/types/jimp.d.ts
@@ -77,7 +77,7 @@ export interface Jimp extends JimpConstructors {
   getExtension(): string;
   distanceFromHash(hash: string): number;
   write(path: string, cb?: ImageCallback): this;
-  writeAsync(path: string): Promise<Jimp>;
+  writeAsync(path: string): Promise<this>;
   rgba(bool: boolean, cb?: ImageCallback): this;
   getBase64(mime: string, cb: GenericCallback<string, any, this>): this;
   getBase64Async(mime: string): Promise<string>;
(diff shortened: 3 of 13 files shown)


crutchcorn:
The previous commit HAS fixed one of the larger issues with the code, but tomorrow I need to have a more thorough discussion with @hipstersmoothie regraurding supported. I'll be sending an email tomorrow morning (PDT) about that

crutchcorn:
Copying from an email I sent to @hipstersmoothie:

There's some questions I had regarding what the supported method of `import` or `require` we want to enable for the TS typings. It's kinda the only question left for the typings AFAIK (Obviously we'll want more users to do testing, but with my most recent commit, the `const A: Jimp = new Jimp('test')` now works (for 3.1 and <3.1)

It seems there's also been a bit of API churn in this way as well:
https://github.com/oliver-moran/jimp/commit/357bd5a34b6f964229bf80c85ef728a8d88c246e
https://github.com/oliver-moran/jimp/commit/7dff2878be3fadc346d15d954bbc9fd7bd749ae7
https://github.com/oliver-moran/jimp/commit/466b65bc6d94f915efe0fa92b30f5f1671456fc9
https://github.com/oliver-moran/jimp/issues/581
https://github.com/oliver-moran/jimp/pull/660
https://github.com/oliver-moran/jimp/issues/785#issuecomment-530964486

So then, the questions are: 
- Is `const Jimp = require('jimp')` allowed? 
(If yes, we'll need to `export = Jimp`)
- If so, should we also be able to `import {PrintableText} from 'jimp'`?
(If yes and the previous question is also yes, we'll have to go back to namespacing the way prior to 660 being merged)
- If we DON'T care about `require` use, do we want to enable:
`import * as Jimp from 'jimp'` or `import Jimp from 'jimp`?
(We can enable both of these depending on the `esModuleInterop` and `allowSyntheticDefaultImports` values of a user's TSConfig)

I personally think we should not allow `const Jimp = require('jimp')` for TS typings and enforce `import Jimp`

Regardless of our choice, we should edit the README. I know how to achieve the types of all of these, but I just need to know which route we're wanting to go. Keep in mind, these answers will impact how we handle our tsconfig.js

crutchcorn:
After the `export`/`import` answers are resolved, I _believe_ (tests witholding) I have TS (all versions) fully working with class usage/non class usage/etc. The tests are now very lengthy and contain every usecase I could come accross.

hipstersmoothie:
>  const Jimp = require('jimp')

This just has to work in plain JS. In TS i'm fine if that doesn't work

crutchcorn:
I've made the ability to use `require('jimp')` and `import * as Jimp from 'jimp` not valid in the TS typings anymore. 
> Aside: I have reason to believe they were both already removed in 0.6.4 without being announced as such, which is why we have users saying their typings did not work on anything higher, but that's another story

As a result, I have updated the docs in the `README` regarding the change. We should probably also mention it SOMEHOW. I don't know if that means a minor bump and warning TS users that 0.8 is no good or if it's just mentioned in a release note that 0.6.4 introduced this/etc, but it's there. I am going to ask the users in the issue thread about testing the most recent published canary with the caviat that neither of these old imports methods work in TS (but work just as they would usually in vanilla JS)

amirburbea:
I assume you know this, but generally if `allowSyntheticDefaultImports` is on, the way to load it should be to use `import jimp from 'jimp'` (otherwise if the flag is off `import * as jimp from 'jimp'`).

Microsoft recommends projects to aim to have the flag on.

crutchcorn:
@amirburbea I actually struggle a lot when it comes to `allowSyntheticDefaultImports` and such. I'm much better at the actual typings themselves. I'd love a review from you on the README.md changes I just made and would love if you'd be able to test both to confirm the validity of my statements.

crutchcorn:
This might mean that the comment here: https://github.com/oliver-moran/jimp/issues/785#issuecomment-532438181

Is not accurate. Again, I'll defer to @amirburbea as I'm unfamiliar with the different import techniques w/ the config settings

crutchcorn:
This is confirmed working by some folks testing from the issue. @hipstersmoothie let's go ahead and merge this, cut a release, then I'll open up a new issue for the problems with the 3.1 typings, then make a PR to fix that

hipstersmoothie:
:rocket: PR was released in v0.8.3 :rocket:

Issue #797, opened 2019-09-18
Uncaught ReferenceError: regeneratorRuntime is not defined


mubaidr:
# Expected Behavior

Jimp browser version should load

# Current Behavior

Jimp fails to load in browser

# Failure Information (for bugs)

index.js?c2d7:3 Uncaught ReferenceError: regeneratorRuntime is not defined

## Steps to Reproduce

`import Jimp from "jimp/es"`

**Screenshots**
NA

## Context

Browser, trying to use jimp in @vue-cli project

- Jimp Version:0.8.2
- Operating System: Windows 10
- Node version: 12.10.0

## Failure Logs
```bash
Uncaught ReferenceError: regeneratorRuntime is not defined
    at eval (index.js?c2d7:3)
    at Module../node_modules/@jimp/utils/es/index.js (app.js:1259)
    at __webpack_require__ (app.js:727)
    at fn (app.js:101)
    at Module.eval (index.js?67ea:1)
    at eval (index.js:1303)
    at Module../node_modules/@jimp/core/es/index.js (app.js:855)
    at __webpack_require__ (app.js:727)
    at fn (app.js:101)
    at eval (index.js?0383:1)
```

crutchcorn:
I'm getting the same problem

crutchcorn:
Are you using TypeScript @mubaidr ?

crutchcorn:
![image](https://user-images.githubusercontent.com/9100169/65362213-6c120b80-dbbb-11e9-9aaa-d19d4da85163.png)

FWIW, I'm using TS, trying to call the `configure` code, and hitting the same problem. However, if I try using the values that TS is clearly MEANING to handle, it is hitting this problem.

crutchcorn:
Adding the following into my TS config has fixed the problem:
```
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
```

I admit total fault for removing this from the README. I'll make a PR to add this back

crutchcorn:
Wait, I'm confused. It seems that I had to add both:
`yarn add -D regenerator-runtime`

As well as the `esModuleInterop` and `allowSyntheticDefaultImports` to work properly. I'm unclear what's exactly happening. Let me see if I can narrow scope of when this occured (please not 0.8, please not 0.8, please)

hipstersmoothie:
Related https://github.com/oliver-moran/jimp/pull/783

crutchcorn:
[...]

PR #799, opened 2019-09-21
Upgrade nearly-all dev deps

crutchcorn:
# What's Changing and Why

Upgrading dev deps, there should be no breaking changes to consumption

The only three dev deps were not updated were:
- `@types/yargs` - seemed to break some type checks on CLI when doing so 
- `ts-node` - seemed to break some things in a build step when doing so, I know some tools rely on this via peer dep, so I assumed not upgrading for now would be best
- `@types/log-symbols` - v3 turns into a shim type as v3 of the library added TS support https://github.com/sindresorhus/log-symbols/releases. However, v3 of the library also ups the required node version, which would mean a breaking change for jimp that I didn't want to introduce until more work on #797 

I also had to run `xo --fix` to fix some things that were added/changed somewhere in the linting process when upgrading `eslint`, `prettier`, I'm unsure

This would drop support for any version of node > 8 for local dev

There seem to be new *warnings* for linting, but no errors

Published PR with canary version: `0.8.5-canary.799.447.0`


crutchcorn:
@hipstersmoothie seems there is some issue with `auto` being released as-of-7.5 or 7.6. Seeing how you contributed to 7.4 I figured I'd give you a heads up before reverting it:

https://circleci.com/gh/oliver-moran/jimp/412?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

hipstersmoothie:
:rocket: PR was released in v0.8.5 :rocket:

crutchcorn:
@hipstersmoothie I don't mean to complain too much, but the CHANGELOG.md seems to have associated this PR as well as #800 to @popinguy 

https://github.com/oliver-moran/jimp/blob/master/CHANGELOG.md#v085-fri-oct-18-2019

(The same on the `Releases` tab)

Issue #803, opened 2019-10-13
TypeScript: Jimp types are not corresponding to Jimp implementation


RomainFallet:
# Expected Behavior

When importing Jimp as described in the doc for a TypeScript project:
```typescript
import Jimp from 'jimp'
```
I should be able to use Jimp directly as is:
```typescript
Jimp
  .read(fileData)
  .then((image: any) => {
    ...
  })
  .catch((err: any) => {
    ...
  })
```

# Current Behavior

The compile step is OK, but there is  **runtime error** saying:
```
TypeError: Cannot read property 'read' of undefined
```

If I use the require-like import syntax:
```typescript
import * as Jimp from 'jimp'
```

and use Jimp as below:
```typescript
Jimp
  .read(fileData)
  .then((image: any) => {
    ...
  })
  .catch((err: any) => {
    ...
  })
```

I then have a **compile error** saying:
```
Property 'read' does not exist on type 'typeof import("/path/to/node_modules/jimp/types/ts3.1/index")'
```

If I disable manually type checking for Jimp:
```typescript
// @ts-ignore
Jimp
  .read(fileData)
  .then((image: any) => {
    ...
  })
  .catch((err: any) => {
    ...
  })
```

There is no **runtime error** and Jimp works as expected.

# Failure Information (for bugs)

## Context
- Jimp Version: 0.8.4
- Operating System: macOS 10.15 Catalina
- Node version: 10.16.3
- NPM Version: 6.11.3
- TypeScript Version: 3.6.3

# Workaround

To use Jimp with types information, I have to import the types separately and bind them to the classic require syntax:

```typescript
import Jimp from 'jimp'
// tslint:disable-next-line: no-var-requires
const jimp: Jimp = require('jimp')
```

And then, use "jimp" instead of "Jimp":
```typescript
jimp
  .read(fileData)
  .then((image: any) => {
    ...
  })
  .catch((err: any) => {
    ...
  })
```

hipstersmoothie:
@crutchcorn

crutchcorn:
I sincerely think we should revert the export type to only export using namespace ala 0.6.4 typings. This doesn't mean we'd revert the 3.1 typings. I don't know enough about tsconfig options to provide an alternative.

crutchcorn:
I have no clue what's happening with @noomly example. I'd have to look further

WeekendMan:
Same for me.
Types are completely fucked.
Documentation is completely fucked.
[...]

PR #820, opened 2019-11-21
Revert exports to match 0.6.4 TS definitions

crutchcorn:
# What's Changing and Why

Many developers using TypeScript have reported that 0.6.4 was the last time that they were able to use the TypeScript definition files. This is because in 0.6.5 the `export` format was modified to more accurately fall inline with ESModule export syntax. 

While TypeScript compiler, with some settings, allow you to handle this export and have the correct results, if the project feels it better to use the same exports as 0.6.4, this is how it would be done without removing the additional types from missing plugins, removing the TS type tests, and more

This would be a breaking change from the current (or even the version 0.6.5) definition files as we no longer export interfaces such as `Bitmap` or others. This is because this `export = Object` does not allow for any other exports to be named in a definition file. Needless to say, _someone_ is not going to be happy with the typings regardless of what we do. I'm just opening this PR per https://github.com/oliver-moran/jimp/issues/803#issuecomment-556872322 as it seems like _this_ is the direction we want to go in

## Tasks

- [x] Add tests
- [ ] Update Documentation
- [x] Update `jimp.d.ts`

### DIFF
--- a/packages/jimp/types/index.d.ts
+++ b/packages/jimp/types/index.d.ts
@@ -3,7 +3,7 @@
 
 declare const DepreciatedJimp: DepreciatedJimp;
 
-export default DepreciatedJimp;
+export = DepreciatedJimp;
 
 /**
  * @deprecated Jimp typings for TS <3.1 are being depreciated. Please upgrade your TypeScript version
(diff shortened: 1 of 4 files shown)


crutchcorn:
Pinging @hipstersmoothie as the _release_ for this version seems to've failed for unrelated reasons

hipstersmoothie:
:rocket: PR was released in v0.9.0 :rocket:

PR #857, opened 2020-03-18
Relax mkdirp dependency to allow newer minimist

Den-dp:
# What's Changing and Why
Fixing https://www.npmjs.com/advisories/1179 by relaxing mkdirp versions range to allow 0.5.2 or 0.5.3 which contains a fix as per https://github.com/isaacs/node-mkdirp/issues/7#issuecomment-600231795

## What else might be affected
Nothing 

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`
- [ ] Add [SemVer](https://semver.org/) Label

Published PR with canary version: `0.9.6-canary.857.681.0`

PR #858, opened 2020-03-18
Fix TypeScript error on 'next'

crutchcorn:
# What's Changing and Why

TS tests seem to be failing in #857. I am unable to recreate this issue on my local machine, but I'm hoping this fixes it on CI (I think my local machine simply isn't getting `next` on TS properly)

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [x] Update `jimp.d.ts`

Published PR with canary version: `0.9.6-canary.858.667.0`

### DIFF
--- a/packages/core/types/jimp.d.ts
+++ b/packages/core/types/jimp.d.ts
@@ -170,14 +170,19 @@ export interface Jimp extends JimpConstructors {
   ): this;
 
   // Functions
-  appendConstructorOption<T extends any[]>(
+  /**
+   * I'd like to make `Args` generic and used in `run` and `test` but alas,
+   * it's not possible RN:
+   * https://github.com/microsoft/TypeScript/issues/26113
+   */
+  appendConstructorOption<Args extends any[], J extends Jimp = this>(
     name: string,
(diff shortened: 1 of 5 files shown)


crutchcorn:
Well golly. I guess it's time to figure out the TS `next` caching issue on my machine to debug further.

hipstersmoothie:
We could also try to limit it to not test @next, but I couldn't find that in the docs

crutchcorn:
We could do that as well, but I'm not able to find a way to test other versions of TS in a range without selecting `next` either. Let me see if I can't fix this - it'd be a shame if TS3.8 came out and we got a new batch of issues, ya know?

crutchcorn:
Apologies for taking so long - I found a bug with TS itself that we can't work around that was impacting some type issues. I decided to edge-case that bug while I was already in the weeds of things. That said, this should work for `next` as well as the other versions. I was able to get the bugs showing on my local machine and they no longer do. Let's hope CI passes

hipstersmoothie:
:rocket: PR was released in `v0.9.6` :rocket:

Issue #689, opened 2019-01-03
Error on using image boxShadow


vinaymuthanna0727:
# Expected Behavior

Need to create an shading/shadow image. 

# Current Behavior

Getting error on using image.boxShadow

# Failure Information (for bugs)

Debugger listening on ws://127.0.0.1:26445/91cb9e8a-85a4-4dcf-b420-043ca9301d64
Debugger attached.
e:\EDrive\shading\index.js:9
        image.boxShadow({ opacity: 0.8, size: 1.2, blur: 10, x: -75, y: -75 });
              ^

TypeError: image.boxShadow is not a function
    at Jimp.<anonymous> (e:\EDrive\shading\index.js:9:15)
    at Timeout._onTimeout (e:\EDrive\shading\node_modules\@jimp\core\dist\index.js:331:25)
    at ontimeout (timers.js:498:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:290:5)
Waiting for the debugger to disconnect...
TypeError: image.boxShadow is not a function
index.js:9
    at Jimp.<anonymous> (e:\EDrive\shading\index.js:9:15)
    at Timeout._onTimeout (e:\EDrive\shading\node_modules\@jimp\core\dist\index.js:331:25)
    at ontimeout (timers.js:498:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:290:5)

## Steps to Reproduce
These are the two file used to create shadow image 

Step1: package.json
       {
  "name": "shading",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "prestart": "npm install",
    "start": "node index.js",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@jimp/custom": "^0.6.0",
    "@jimp/plugin-blur": "^0.6.0",
    "@jimp/plugin-resize": "^0.6.0",
    "@jimp/plugin-shadow": "^0.6.0",
    "@jimp/test-utils": "^0.6.0",
    "@jimp/utils": "^0.6.0",
    "core-js": "^2.6.1",
    "jimp": "^0.6.0"
  }
}

step2: index.js 

var jimp = require('jimp');
var inputPath = "E:/EDrive/Images/2016_12_07_18_27_16--Project002/img.png";
[...]

PR #863, opened 2020-03-27
Added missing plugins to the types

crutchcorn:
# What's Changing and Why

As mentioned by @SaliZumberi during a discussion on #689, I noticed we were missing a few plugins from the core typings for typings versions 3.1+. This PR adds those plugins and the typings associated

This should be a minor release as it doesn't break any functionality

## Tasks

- [x] Add tests
- [ ] Update Documentation
- [x] Update `jimp.d.ts`


hipstersmoothie:
:rocket: PR was released in `v0.9.7` :rocket:

Issue #833, opened 2019-12-29
WebAssembly version for speed.


intervalia:
I was wondering if you were considering making a WebAssembly (WASM) version of your library.

WebAssembly has been in Node since version 8. Your code could run substantially faster if ported to WASM and it would be compatible with all LTS versions of Node

Just a suggestion that could help improve your library.

elad-yosifon:
👍

forivall:
I would recommend porting the code to assemblyscript.

forivall:
WHOOPS i started working on implementing this.

ChristianUlbrich:
Why should a WASM version be significantly faster, than the JavaScript version? Just for context, have a look at: https://colineberhardt.github.io/wasm-mandelbrot/#WebAssembly

Compare the JS version of the Mandelbrot to the AssemblyScript one:
* https://github.com/ColinEberhardt/wasm-mandelbrot/blob/master/js/mandelbrot.js
* https://github.com/ColinEberhardt/wasm-mandelbrot/blob/master/assemblyscript/mandelbrot.ts 

_Simply_ translating code to assemblyscript will not likely give you a speed boost. While assemblyscript claims to have a _optimized native Math implementation_ I could not find any real world numbers.

I do not want to discourage you, but I really doubt your initial statement:
> Your code could run substantially faster if ported to WASM

For context have a look at the following articles:
* https://stackoverflow.com/questions/48173979/why-is-webassembly-function-almost-300-time-slower-than-same-js-function
* http://link.crwd.fr/45ZL#https://blog.acolyer.org/2017/09/18/bringing-the-web-up-to-speed-with-webassembly/?utm_campaign=crowdfire&utm_content=crowdfire&utm_medium=social&utm_source=twitter
* https://discourse.threejs.org/t/porting-three-js-to-webassembly-with-assemblyscript/11629/6

hipstersmoothie:
I've never really used wasm myself. but it seems like the best way to use wasm in jimp would either be through:

- using wasm-ified PNG, JPEG, etc 
- Building a new version of jimp (probably a new project) written on top of `ImageMagick` but `sharp` would probably be the one to do that. 

This could also get ridiculously large and not be suited for browsers anymore

forivall:
@ChristianUlbrich yeah, i read the mandelbrot comparison when i was working on this, and the results i've found has concluded that there isn't a huge amount of speed boost. I'm not actively working on the PR, but it was still worth the investigation.

[...]

PR #874, opened 2020-04-13
WIP: asm plugin for wasm-based speedups via assemblyscript

forivall:
# What's Changing and Why

Adding a plugin to run operations in wasm. Fix for #833 .

Oh boy. A lot might be. This is very proof of concept right now and i'm tired and i can explain more at a later time.

Also, since I prefer writing typescript, i added the babel typescript preset to the root config.

## Tasks

- [ ] Add code to reduce data copies between the wasm memory and js memory.
- [ ] Determine the best way to override other method's internals to use wasm-mode
  - Check out the changes in `blit` for how i'm currently patching the asm code in.
- [ ] Port `composite`
- [x] Port `blit`
- [ ] Port `blur`
- [ ] Port `circle`
- [ ] Port `color`
- [ ] Port `contain`
- [ ] Port `cover`
- [ ] Port `crop`
- [ ] Port `displace`
- [ ] Port `dither`
- [ ] Port `fisheye`
- [ ] Port `flip`
- [ ] Port `gaussian`
- [ ] Port `invert`
- [ ] Port `mask`
- [ ] Port `normalize`
- [ ] Port `print`
- [ ] Port `resize`
- [ ] Port `rotate`
- [ ] Port `scale`
- [ ] Port `shadow`
- [ ] Port `threshold`
- [ ] Port `phash`
- [ ] Port some formats?
- [ ] Try to make it work in the browser
- [ ] Add tests
- [ ] Add benchmarks
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`


hipstersmoothie:
Will this enable use to use other wasm-ified things easily? Like at some point it would be amazing to include the official PNG library to do PNGs instead of png.js

forivall:
Maybe? At the very least, for external wasmified things, we would have an established pattern for loading wasm. This is really my first adventure in building wasm things myself, but it might be possible to share the memory between different wasm binaries. I'd have to look into that. The main thing to consider when transitioning to wasm is that for ideal performance, you should avoid copying memory in and out of the wasm context. it is a lot faster than moving in and out of a native node module context (i think), but it is still something to consider.

forivall:
Something i stumbled upon that i want to note here: opencv's documentation uses jimp to load images, so an opencv based plugin for speedups also could be possible. https://docs.opencv.org/master/dc/de6/tutorial_js_nodejs.html

hipstersmoothie:
@forivall it's been a long time but if you're still up for this it seems interesting. closing for now

PR #879, opened 2020-04-20
Simplify and fix flip

### LINKED ISSUE #829: Flip / Mirror with BOTH vertical and horizontal parameters as true adds margin to image bounds
# Expected Behavior

Calling flip() or mirror() on an image with both vertical and horizontal parameters as true should flip the image vertically and horizontally, without introducing artifacts to the flipped image.

# Current Behavior

Calling flip (or mirror) on an image with both vertical and horizontal parameters set to true adds a small white (2-3 pixels) margin around the border of the final image. 

n/a -- see steps to reproduce

## Steps to Reproduce

Using this code snippet: 

``` ts
      Jimp.read('some/local/image/url/from/assets/dir').then(img => {
        const topLeft = img.clone().crop(0, 0, img.getWidth() / 2, img.getHeight() / 2).flip(true, true);
        const topLeft2 = img.clone().crop(0, 0, img.getWidth() / 2, img.getHeight() / 2).flip(true, false).flip(false, true);
}
```
See screenshots below. 

Left is the source image. I'm extracting quadrants from the image using crop, then flip each quadrant vertically and horizontally. 
The center image is topLeft (method above). 
The right-most image is topLeft2 (method above). 

I'm using a thick CSS border around each image to emphasize the white space issue I'm seeing. Calling the first method chain for "topLeft" introduces that white margin. Specifically, calling flip(true, true), results in the artifact. When calling flip(true, false).flip(false, true), I get the desired result, with no white "margin" around the image. 

I've used three separate images, and this occurs on each of them. 

**Screenshots**

image

## Context

- Jimp Version: "jimp": "^0.9.3", (browser version, typescript, with Angular 8
- Operating System: Windows 10
- Node version: 10.16.3

## Failure Logs
none

skalee:
# What's Changing and Why

- Tests for Flip plugin are added.
- A code branch in Flip plugin is removed, as it was broken and unnecessary.

In the result of the latter:

- Computational complexity is the same or better.
- Flip plugin no longer depends on Rotate plugin.
- Implementation is simpler, shorter and has less bugs.

Fixes #829. API is unaffected.

## Tasks

- [x] Add tests
- [x] Update Documentation — does not apply
- [ ] Update `jimp.d.ts` — I suppose irrelevant, as the API is unaffected

### DIFF
--- a/packages/plugin-flip/src/index.js
+++ b/packages/plugin-flip/src/index.js
@@ -15,11 +15,6 @@ function flipFn(horizontal, vertical, cb) {
       cb
     );
 
-  if (horizontal && vertical) {
-    // shortcut
-    return this.rotate(180, true, cb);
-  }
-
   const bitmap = Buffer.alloc(this.bitmap.data.length);
   this.scanQuiet(0, 0, this.bitmap.width, this.bitmap.height, function(
     x,
--- a/packages/plugin-flip/test/flipping.test.js
+++ b/packages/plugin-flip/test/flipping.test.js
@@ -1,0 +1,99 @@
+import { Jimp, mkJGD } from '@jimp/test-utils';
+
+import configure from '@jimp/custom';
+
+import flip from '../src';
+
+const jimp = configure({ plugins: [flip] }, Jimp);
+
+describe('Flipping plugin', () => {
+  it('can flip horizontally', async () => {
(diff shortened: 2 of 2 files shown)


hipstersmoothie:
:rocket: PR was released in `v0.10.3` :rocket:

Issue #694, opened 2019-01-16
Browser version requires missing files.


wallabra:
# Expected Behavior

Using the browser version of JIMP, e.g. when running Browserify on a Node.JS package that requires JIMP, should be clean, as it should contain all of the JIMP code in a single file.

# Current Behavior

Browserify (most likely the browser too, if you use the browser version directly in the browser) will error due to erroneous calls to `require()`.

# Failure Information (for bugs)

The error complains about `./src/converter`, which simply doesn't exist in `jimp/browser` (where `jimp` is the `node_modules` subfolder where `jimp` is installed, when installed from `npm`).

## Steps to Reproduce

I haven't tried to use JIMP directly in the browser, but here's how to reproduce this error the way I noticed it.

1. Create a simple Node.JS project.
2. Install JIMP as a dependency (`npm i --save jimp`).
3. Create an `index.js` file with the following contents:

```
const JIMP = require('jimp');
```

4. Run `browserify index.js`.

## Context

- Jimp Version: JIMP 0.6.0.
- Operating System: Linux Mint 19 "Tara" (Ubuntu fork), kernel Linux v4.20.
- Node version: Node.JS v8.10.0.

## Failure Logs

```
> nodemips@0.1.1 build /home/gustavo6046/Source code/nodemips
> browserify -r ./index.js:mips | babel --minified -f minif.tmp | uglifyjs >browser/mips.js

Error: Cannot find module './src/converter' from '/home/gustavo6046/Source code/nodemips/node_modules/jimp/browser/lib'
    at /home/gustavo6046/Source code/nodemips/node_modules/browser-resolve/node_modules/resolve/lib/async.js:55:21
    at load (/home/gustavo6046/Source code/nodemips/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
    at onex (/home/gustavo6046/Source code/nodemips/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
    at /home/gustavo6046/Source code/nodemips/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:152:21)
```

Gustavo6046:
I've noticed that simply removing the `browser` field from JIMP's `package.json` seems to fix the issue (although I haven't tested the output in a browser environment yet).

hipstersmoothie:
Do you think should be browserifying itself? Or should it be left to users

[...]

PR #918, opened 2020-07-26
Remove 'browser' field from main jimp package's package.json. Fixes:#694

Gustavo6046:
# What's Changing and Why

The `browser` tag in `packages/jimp/package.json` is being removed. It interferes with Browserify's dependency resolution, breaking it and ironically rendering Jimp unusable in Browserify. Meanwhile, building a single JavaScript file for manual inclusion in a webpage remains trivial, as is distributing it, using Browserify.

Nothing, which is why I made the decision to simply remove the disruptive field rather than replacing it with something else. Distributing the JavaScript file for clientside (e.g. HTML) inclusion of Jimp is done separately from `npm`, because browsers (and other client-side JavaScript applications) don't use `npm`, which is for server-side applications (i.e. Node.js). That, however, is a concern beyond that of this specific pull request.

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`

### DIFF
--- a/packages/jimp/package.json
+++ b/packages/jimp/package.json
@@ -4,7 +4,6 @@
   "description": "An image processing library written entirely in JavaScript (i.e. zero external or native dependencies)",
   "main": "dist/index.js",
   "module": "es/index.js",
-  "browser": "browser/lib/jimp.js",
   "types": "types/index.d.ts",
   "typesVersions": {
     ">=3.1.0-0": {


hipstersmoothie:
:rocket: PR was released in `v0.15.0` :rocket:

Gustavo6046:
Can I see what error this pull request produces in CI? I would like to try and fix it.

hipstersmoothie:
It was all good. published now

Gustavo6046:
Ah, alright. Thank you.

PR #933, opened 2020-08-28
upgrade jpeg-js dependency

Chupsy:
# What's Changing and Why
This PR relates to the new release of `jpeg-js` (more info [here](https://github.com/eugeneware/jpeg-js/pull/79)). Some jpeg files were not properly handled by `jpeg-js` therefore are refused by `jimp`. Objective is to upgrade the library so that these files are accepted on `jimp`.
Yarn lockfile seemed to be out-of-date, so adding this dependency updated the lockfile for all missing packages. It should not be a problem. All tests pass.
## Tasks

- [ ] Add tests **Unnecessary**
- [ ] Update Documentation **Unnecessary**
- [ ] Update `jimp.d.ts` **Unnecessary**

### DIFF
--- a/packages/type-jpeg/package.json
+++ b/packages/type-jpeg/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "@babel/runtime": "^7.7.2",
     "@jimp/utils": "link:../utils",
-    "jpeg-js": "^0.4.0"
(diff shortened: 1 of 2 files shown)


hipstersmoothie:
:rocket: PR was released in `v0.16.1` :rocket:

PR #951, opened 2020-10-07
use a consistent text layouting algorithm

iliazeus:
# What's Changing and Why

In the `print` plugin, there are two places where the text is split in lines no longer than `maxWidth`. They all use slightly different algorithms, which produces inconsistent results. This PR fixes that by using the `splitLines()` function in both places.

While it is, technically, a breaking change, I am fairly sure that nothing breaks because the behavior was both unexpected and undocumented.

## Tasks

- [x] Add tests
- [x] Update Documentation (not needed)
- [x] Update `jimp.d.ts` (not needed)

### DIFF
--- a/packages/plugin-print/package.json
+++ b/packages/plugin-print/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jimp/plugin-print",
-  "version": "0.16.1",
+  "version": "0.16.2",
   "description": "print an image.",
   "main": "dist/index.js",
   "module": "es/index.js",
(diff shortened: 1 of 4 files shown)


packages/plugin-print/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jimp/plugin-print",
-  "version": "0.16.1",
+  "version": "0.16.2",
```
hipstersmoothie, line 3: Revert this, it happens automatically

hipstersmoothie (approved):

hipstersmoothie:
closed in favor of https://github.com/jimp-dev/jimp/pull/1133

iliazeus:
Sorry, was on vacation, only just got to my PC. Thanks for accepting my contribution!

PR #979, opened 2021-01-12
upgrade jpeg-js dependency

AT41:
# What's Changing and Why
jpeg-js has fixed a [new issue](https://github.com/eugeneware/jpeg-js/issues/82). Pixel 4 XL phones took malformed .jpegs, which raised exceptions in previous versions of jpeg-js.

Yarn.lock seemed out of date, thus the various changes to yarn.lock. There also seemed to be an old version of jpeg-js that got automatically removed, however tests all passed. No other packages have dependencies to older versions of jpeg-js so removing previous versions should be fine.

## Tasks

- [ ] Add tests **Unnecessary**
- [ ] Update Documentation **Unnecessary**
- [ ] Update `jimp.d.ts` **Unnecessary**
- [ ] Add [SemVer](https://semver.org/) Label **Unnecessary**

### DIFF
--- a/packages/type-jpeg/package.json
+++ b/packages/type-jpeg/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "@babel/runtime": "^7.7.2",
     "@jimp/utils": "link:../utils",
-    "jpeg-js": "0.4.2"
+    "jpeg-js": "0.4.3"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


AT41:
The Node8 checks seem to fail because gifs will occasionally take more than 2000ms. I've rerun CircleCI multiple times and have differing test results on the same branch commit.

Screenshot from 2021-01-13 11-55-24

PR #1020, opened 2021-06-11
fix commonjs require

fratzinger:
# What's Changing and Why

This little change is for js users who use commonjs-modules.
before the fix: `const Jimp = require('jimp').default;`
after the fix: `const Jimp = require('jimp');`

The esm module isn't affected. Other popular libraries use this fix as well.

nothing

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`


Steve-Mcl:
Will `const Jimp = require('jimp').default` continue to work in existing node js projects?

If not, will this be a major semver change?

fratzinger:
Good point. `.default` will continue working.

hipstersmoothie:
:rocket: PR was released in [`v0.17.4`](https://github.com/jimp-dev/jimp/releases/tag/v0.17.4) :rocket:

PR #1035, opened 2021-07-30
fix: scaleTofit factor calc fixed

zouyixiong:
# What's Changing and Why

const w = 2120, h = 1280
const bitmap = { width: 42, height: 37 }

const factorW = w / bitmap.width = 2120 / 42 = 50.47
const factorH = h / bitmap.height = 1280 / 37 = 34.59
In this case , I think factorW is the right factor.

old version : 
w / h  = 2120 / 1280 = 1.65625
bitmap.width / bitmap.height = 42 / 37 = 1.135
so factor = factorH

no

## Tasks

- [ no required ] Add tests
- [ no required] Update Documentation
- [ no required ] Update `jimp.d.ts`
- [ no required ] Add [SemVer](https://semver.org/) Label

### DIFF
--- a/packages/plugin-scale/src/index.js
+++ b/packages/plugin-scale/src/index.js
@@ -51,10 +51,7 @@ export default () => ({
       mode = null;
     }
 
-    const f =
-      w / h > this.bitmap.width / this.bitmap.height
-        ? h / this.bitmap.height
-        : w / this.bitmap.width;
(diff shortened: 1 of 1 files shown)


hipstersmoothie:
I just tested this and on:

1. a 512x512 image
2. `image.scaleToFit(200, 100)`

With the current code the image is 100x100 with your code it's 200x200 which is wrong

PR #1049, opened 2021-09-23
Enable importing SVG

### LINKED ISSUE #118: SVG support
Is SVG support planned or was it left out on purpose? Just wondering as I'm looking into dropping `node-gm` in favour of a library that does not depend on external binaries.

josh-hemphill:
closes #118
# What's Changing and Why
Adding a means to import SVG files and render them.
This is just an initial proof of concept at the moment, because the workflow for allowing pre-render manipulation of SVG is a lot more complicated than any other plugins or type support currently implemented, and needs some more thought put into it.
In it's current state it also node-only since its imports of node polyfills for canvas, xml, and fetch are not dynamically applied by platform. And a few method calls would have to change depending on the platform as well.

Nothing I can think of.

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`


hipstersmoothie:
I'm not sure that jimp support svg makes all the much sense. I can see that case for decode it into a bitmap. But encoding seems like it's impossible. jimp lets you convert from one format to another. converting a raster image to vector seems like a much bigger thing than jimp sets out to solve

PR #1154, opened 2023-02-05
Change some exports to move towards more ESM compatibility

danielholmes:
# What's Changing and Why

There's some parts of the source code that still use commonJS export style. This change makes a more consistent code base and once completely done, open opportunities to simplify and improve the compile steps (babel, browserify, etc).

Note: There are still 1 locations with CJS exports, one i being handled in [another PR](https://github.com/jimp-dev/jimp/pull/1153), the other is the [core requests module](https://github.com/jimp-dev/jimp/blob/8c0d0f3b8232d08acac06e9d7bfc5789489c6e1c/packages/core/src/request.js) which requires a larger discussion (PR forthcoming).

It's a pretty simple change. The tests run for me and I can't imagine it breaking anything

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`

### DIFF
--- a/packages/core/src/modules/phash.js
+++ b/packages/core/src/modules/phash.js
@@ -179,4 +179,4 @@ function applyDCT(f, size) {
   return F;
 }
 
-module.exports = ImagePHash;
+export default ImagePHash;
--- a/packages/plugin-resize/src/modules/resize.js
+++ b/packages/plugin-resize/src/modules/resize.js
@@ -545,4 +545,4 @@ Resize.prototype.generateUint8Buffer = function (bufferLength) {
   }
 };
 
-module.exports = Resize;
+export default Resize;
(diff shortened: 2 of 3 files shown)


hipstersmoothie (approved):

hipstersmoothie:
Hmm I don't know what the actions aren't running

danielholmes:
@hipstersmoothie I found some discussion about this in the actions/checkout repo:

https://github.com/actions/checkout/issues/298

[One suggested fix is](https://github.com/actions/checkout/issues/298#issuecomment-820748989):

```
token: ${{ secrets.MY_TOKEN || github.token }}
```

in the workflow

hipstersmoothie:
ah. I'm not sure that line was necessary. If it is thanks for the heads up!

hipstersmoothie:
These other failures are for the canary release. I want to make running that part something I can opt into via label or something. So for now it will only work for people with access to the repo

Issue #1129, opened 2023-02-03
Get karma browser tests running again


hipstersmoothie:

nopeless:
@hipstersmoothie on it

hipstersmoothie:
Oh wait!

hipstersmoothie:
Before you do this I think you should look into using this project

https://twitter.com/HipsterSmoothie/status/1622298963486572544

It seems a bit newer and support using playwright which would be really nice to use.

hipstersmoothie:
It might also be easier to get playwright running on CI than karma

hipstersmoothie:
we use vitest browser mode now

PR #1157, opened 2023-02-05
Fix-karma

nopeless:
Karma has been fixed


nopeless:
#1129

hipstersmoothie:
Could you try to remove all the browserify modules? I think there is a webpack-karma that might be better. Working on fixing CI right now for PRs

nopeless:
> Could you try to remove all the browserify modules? I think there is a webpack-karma that might be better. Working on fixing CI right now for PRs

Sounds good.

hipstersmoothie:
Here is where to turn the tests back on https://github.com/jimp-dev/jimp/blob/master/.github/workflows/build.yml#L69

nopeless:
@hipstersmoothie are we going to change browser testing framework in the TS release/next branch?

hipstersmoothie:
I think keeping the ability to run the tests in the browser is a good idea. 

I think karma is kinda old and we could replace it with [this](https://github.com/jimp-dev/jimp/issues/1129#issuecomment-1418286649) and playwright and get something better

Since I want to test in browser mocha will probably stay but I'm working on switching the assertion library from should to expect.

hipstersmoothie:
The TS refactor shouldn't effect any of this

hipstersmoothie:
Well expect doesn't work in browser </3

nopeless:
> The TS refactor shouldn't effect any of this

yes but point is why not change testing framework when changing code as well

nopeless:
@hipstersmoothie Moving to webpack is causing issues with `package/test-utils` as apparantly Jimp had some kind of hack for loading tests

hipstersmoothie:
Id try the newer tool before any further getting karma working. 

If it makes it simpler you can also assume for the browser tests that jimp has already been bundles and exists in packages/jimp/browser/lib

nopeless:
@hipstersmoothie As some pointed out, we still have to maintain it for some time. We should still use Karma for the JS version of Jimp

nopeless:
@hipstersmoothie With that being said, even if we go to playwright I belive I would still have to debug this `getTestDir` in `test-utils`

nopeless:
Fixes #1129

nopeless:
Issue found. esmodules don't have a prototype

Resorting to .default should be the fix

nopeless:
image
Another reason why I advocate for a full rewrite. This is not maintainable behavior

hipstersmoothie:
@nopeless I may have a PR that makes this easier

nopeless:
If you are talking about #1163, it doesn't fix the webpack issue. The conversion from CJS to ESM breaks tests as well

hipstersmoothie:
https://github.com/jimp-dev/jimp/pull/1163

hipstersmoothie:
What/how are you web packing? The library itself already bundles fine we should be able to use that bundle to run tests with in a browser.

> https://github.com/jimp-dev/jimp/pull/1157#issuecomment-1418323259

Could you file an issue for this and propose what you think it should be?

nopeless:
```js
import { Jimp, mkJGD } from "@jimp/test-utils";

describe("Compare image difference", () => {
  let imgs = [
    Jimp.read(mkJGD("2222", "4444", "6666", "8888")),
```

@hipstersmoothie It seems that the Jimp here is the esmodule Jimp itself not Jimp class. That is what's causing the issue

hipstersmoothie:
Ahhh I see

nopeless:
> Ahhh I see

I currently have a hack where on `jgd-wrapper` I unwrap Jimp to Jimp.default when Jimp is esmodule but I believe it is ultimately up to you to decide how you want to resolve this issue

hipstersmoothie:
Ok so if you're trying to get karma to still work I think we should stop that

The steps I see for running the tests in browser:

1. Install this tool https://modern-web.dev/docs/test-runner/overview/
2. Configure it to match our tests
3. probably will need to use https://modern-web.dev/docs/dev-server/plugins/esbuild/
4. configure it to use playwright

nopeless:
> Ok so if you're trying to get karma to still work I think we should stop that
> 
> The steps I see for running the tests in browser:
> 
> 1. Install this tool https://modern-web.dev/docs/test-runner/overview/
> 2. Configure it to match our tests
> 3. probably will need to use https://modern-web.dev/docs/dev-server/plugins/esbuild/
> 4. configure it to use playwright

I am almost certain that this isn't a Karma issue. (Unless you are talking about the esoteric loading system that is in place for karma at the moment. That is certainly an issue that this PR will fix)

If, however, configuring esbuild and an alternative testing system implies removing the current one, I am definitely up for that. For now, I don't think we need more changes other than using webpack.

hipstersmoothie:
Yeah I def wanna remove karma entirely. mocha + expect are here to stay but karma I want gone

nopeless:
> Yeah I def wanna remove karma entirely. mocha + expect are here to stay but karma I want gone

Well, we can do that in another PR. I want Karma at least running so it can run against other fixes while the test migration is in progress

hipstersmoothie:
Sounds like a good plan!

nopeless:
Apparantly the newest build updated yarn.lock for packages as well since I had updated test-utils

With that being said, why is test-utils in src in the first place

packages/test-utils/src/index.js
```diff
@@ -13,7 +13,8 @@ export function hashForEach(hash, func) {
 }
 
 export function getTestDir(dir) {
-  const testRE = /\/[^/]+\.test\.js($|\?.*)/;
+  // Legacy webpack support. I do not know the exact purpose of this file
```
hipstersmoothie, line 16: yeah I have no clue either lol. We'll probably need to revisit when moving away from karma

packages/test-utils/src/jgd-wrapper.js
```diff
@@ -38,8 +38,28 @@ function configureJimp() {
   });
 }
 
-const Jimp =
-  typeof window !== "undefined" && window.Jimp ? window.Jimp : configureJimp();
+/**
+ * The ultimate hack
```
hipstersmoothie, line 42: Could you explain what you think is going on?

hipstersmoothie:
@nopeless you can update the lock file just by running `yarn`

hipstersmoothie:
Organizing packages like this is pretty standard for JS monorepos

packages/test-utils/src/jgd-wrapper.js
nopeless: Essentially this is just recursively unwrapping Jimp to get Jimp class. The reaason why I made it recursive is because I do not know what kind of updates you will make so this is a dirty future proof hack

hipstersmoothie: Could you add that in the comment?

nopeless:
The lock file is already updated. I just found it weird that tests loaders are coupled with the versioning. I understand why tests are part of the versions but the loaders should be minimal and not versioned; A gotcha moment for me

nopeless:
@hipstersmoothie I see wht you mean

hipstersmoothie:
whoops this closed cause I renamed master

PR #1165, opened 2023-02-06
Switch to fetch for url requests

danielholmes:
# What's Changing and Why

[Core/request](https://github.com/jimp-dev/jimp/blob/8c0d0f3b8232d08acac06e9d7bfc5789489c6e1c/packages/core/src/request.js) is the last source file using commonJS exports. It's not a simple syntax change though since it uses a feature of CJS not available in ESM - conditional exports. Specifically, if in a browser or electron environment then the global `XMLHttpRequest` is being used, but in a node environment the node only package [`phin`](https://www.npmjs.com/package/phin) is being used.

This is one of the hard blockers for ESM based tool compatibility (e.g. vite and rollup). See [my example repo](https://github.com/danielholmes/jimp-vite) for a playground trying to get this working.

NodeJS has `fetch` built in from v18 onwards and [browser compatibility for fetch is now very high](https://caniuse.com/fetch). I propose a move to fetch instead of the conditional requires.

I went back and forth about whether to use fetch and rely on the library users to install a polyfill if needed or to include a polyfill within jimp. Without the polyfill I think a lot of projects will break and I think that should maybe be done in a "big" release (like `1.0`), but let me know what you think.

`loadFromURL` can take a set of `options`. When in the node environment all those options are passed to `phin`. Any options used to pass to phin that are not supported by the fetch API will stop working.

## Tasks

- [x] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`

### DIFF
--- a/packages/core/test/request.test.js
+++ b/packages/core/test/request.test.js
@@ -35,18 +35,18 @@ const httpHandler = (req, res) => {
 };
 
 describe("request", () => {
-  const httpServer = http.createServer(httpHandler);
-  before(() => {
-    httpServer.listen(5136);
-  });
-
-  after(() => {
-    httpServer.close();
-  });
-
   if (typeof window !== "undefined" && typeof window.document !== "undefined") {
     xit("Not testing requests in browser");
   } else {
+    const httpServer = http.createServer(httpHandler);
+    before(() => {
+      httpServer.listen(5136);
+    });
+
+    after(() => {
+      httpServer.close();
+    });
+
     it("loads standard response", async () => {
       await jimp.read("http://localhost:5136/corrected.png");
     });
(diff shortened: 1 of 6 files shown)


packages/core/test/request.test.js
```diff
@@ -33,26 +34,33 @@ const httpHandler = (req, res) => {
...
-describe("redirect", () => {
+describe("request", () => {
+  const httpServer = http.createServer(httpHandler);
+  before(() => {
+    httpServer.listen(5136);
+  });
+
+  after(() => {
+    httpServer.close();
+  });
```
hipstersmoothie, line 45: I think the still need to be in the if to make the browser tests runnable

hipstersmoothie (changes requested):

packages/core/test/request.test.js
danielholmes: Whoops that was a mistake. Fixed now. Thanks.

hipstersmoothie (approved):

PR #1198, opened 2023-02-09
Yarn Berry

crutchcorn:
This PR adds in Yarn Berry as the repository's package management system to replace Yarn 1.

It also bumps Lerna to v6, as v3 does not support `workspace:` modifiers used by Yarn Berry

Closes #1196

<details>
  <summary>📦 Published PR as canary version: <code>0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @jimp/core@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/custom@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install jimp@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-blit@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-blur@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-circle@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-color@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-contain@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-cover@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-crop@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-displace@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-dither@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-fisheye@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
  npm install @jimp/plugin-flip@0.22.6--canary.1198.9324a88b87a1fc38b528af47ca707d97593dddaa.0
[...]

PR #1199, opened 2023-02-09
Update Type Tests to use Vitest

crutchcorn:
> This PR builds on top of #1198, please merge that first

This PR's overarching goal was to replace `dts-lint` with Vitest for Type Testing, as `dts-lint` is deprecated now.

However, to accomplish this goal cleanly, I had to:

- [x] Updated how GitHub actions calls type tests
- [x] Made lerna call `test:ts` for all packages with the script name, not just the two hardcoded packages
- [x] Add a shared TypeScript configuration in a new `@jimp/config` internal package (to add in missing `tsconfig.json` files without duplicating contents. I plan on expanding this internal package for ESLint, test configs, and more later on)
- [x] Add missing `@jimp/core` dev dependency for all plugin packages (to avoid `any` typings)


hipstersmoothie:
Done ✅

PR #1205, opened 2023-02-22
Export ColorActionName enum

### LINKED ISSUE #987: Image `saturate` / `desaturate` appear on the doc, but not on the module...
# Expected Behavior
Hi, I would like to reduce or increase my image saturation, as specified in the doc [here](https://github.com/oliver-moran/jimp/tree/master/packages/jimp#advanced-usage), but the module doesn't have those options.  
My issue refer to another one, previously closed: https://github.com/oliver-moran/jimp/issues/968

# Current Behavior
`Type '"saturate"' is not assignable to type 'ColorActionName'.` because `saturate` or `desaturate` (and `lighten`, `darken`) doesn't exist on `ColorActionName`.

\-\-\-

## Steps to Reproduce
Try to change the saturation of an image
```js
image.color([{ apply: 'saturate', params: [10] }]);
```
As you can see on the screen, even Intellisense doesn't propose me `saturate`, `desaturate`...
**Screenshots**  
Intellisense

## Context
- Jimp Version: 0.16.1
- Operating System: macOS Big Sur 11.2
- Node version: v14.15.1

## Failure Logs
```
$ npm start
>>> src/helpers/FilterManager.ts(95,17): error TS2322: Type '"saturate"' is not assignable to type 'ColorActionName'.
```

daniellwdb:
# What's Changing and Why
This PR fixes #987 by exporting an enum with all possible values for the `color` method's apply parameters. I chose to use an enum because this makes refactoring easier compared to having plain strings. Originally the `ColorActionName` type was missing some values.

This is a breaking change for projects using TypeScript because they would have to reference the enum directly and use their values. 

Some possible fix for backwards compatibility (>TS4.1):
```diff
++ type ColorActionNameValue = `${ColorActionName}`

type ColorAction = {
-- apply: ColorActionName;
++ apply: ColorActionName | ColorActionNameValue;
```
If older versions need to be supported I can try to look into it.

## Tasks

- [x] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`

### DIFF
--- a/packages/jimp/types/test.ts
+++ b/packages/jimp/types/test.ts
@@ -1,4 +1,5 @@
 import * as Jimp from "jimp";
+import { ColorActionName } from "@jimp/plugin-color";
 
 const jimpInst: Jimp = new Jimp("test");
 
@@ -11,6 +12,7 @@ jimpInst.displace(jimpInst, 2);
 jimpInst.shadow((err, val, coords) => {});
 jimpInst.fishEye({ r: 12 });
 jimpInst.circle({ radius: 12, x: 12, y: 12 });
+jimpInst.color([{ apply: ColorActionName.SATURATE, params: [90] }]);
 // $ExpectError
 jimpInst.PNG_FILTER_NONE;
 
--- a/packages/plugin-color/index.d.ts
+++ b/packages/plugin-color/index.d.ts
@@ -1,16 +1,22 @@
 import { ImageCallback } from "@jimp/core";
 
-type ColorActionName =
-  | "mix"
-  | "tint"
-  | "shade"
-  | "xor"
-  | "red"
-  | "green"
-  | "blue"
-  | "hue"
-  | "lighten"
-  | "darken";
+export enum ColorActionName {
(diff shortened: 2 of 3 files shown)


packages/jimp/types/test.ts
```diff
@@ -1,4 +1,5 @@
 import * as Jimp from "jimp";
+import { ColorActionName } from "@jimp/plugin-color";
```
hipstersmoothie, line 2: I'm not sure this would actually work in runtime code. Our TS is .d.ts only so it doesn't effect the actual JS exported by the package so the enum wouldn't exist.

I think we could keep the `ColorActionName` type the same but export an object from actual JS so this works as expected

hipstersmoothie (changes requested):

packages/jimp/types/test.ts
daniellwdb: Thanks for spotting!

hipstersmoothie (approved):

PR #1209, opened 2023-02-25
use matrix rotate, if rotate angle is multiple of 90 degrees

black-turtle:
Closes https://github.com/jimp-dev/jimp/issues/1101

# What's Changing
Changing rotate algorithm if angle is multiple of 90 degree. For other rotate angle previous algorithm will be used.

## Why Changing
- For `rotate(90)` or `rotate(-90)`, expected new width and height is `width=oldHeight` and `height=oldWidth`. However, existing  algorithms is adding an extra pixel to the rotated image.
- This algorithm should be faster than previous one. Time complexity should be O(w*h).

## Tasks
- [x] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`

## Release Notes

Fixes a bug where rotating by 90 degrees introduced extra pixels


hipstersmoothie (approved):

black-turtle:
Hi @hipstersmoothie 
Most probably we also need to change the Major version. As rotate behavior is changing. Do you want me to change the Major version in package.json file?

hipstersmoothie:
How would this be a breaking change? It seems to fix a bug.

black-turtle:
OK, sounds good. No version change needed then.

PR #1222, opened 2023-04-18
chore: upgrade gifwrap

Marsup:
# What's Changing and Why

I've upgraded the dependency on gifwrap because of this fix https://github.com/jimp-dev/gifwrap/pull/26.

It seems to bring an update of `image-q`, I'm not sure if it's important.

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`


Marsup:
@hipstersmoothie would it be possible to release a new version of jimp with this change please?

PR #1237, opened 2023-05-18
avoid relying on __dirname global constant

### LINKED ISSUE #1235: __dirname err - module build of jimp
IT shoud do No error when using bundlers like Vite

 **__dirname is not defined in ES module scope error**

on this line:
const dir=process.env.DIRNAME||`${__dirname}/../`,print=()=>({constants:{measureText,measureTextHeight,FONT_SANS_8_BLACK:

im using typescript 4.9
syntethicdefaults import true 
but importing es module version 'jimp/es'

1. create importjimp.ts with following:
```
 import configure from "@jimp/custom";
import types from "@jimp/types";
import plugins from "@jimp/plugins";
export default configure({
  types: [types],
  plugins: [plugins]
});
```

2. import jimp from importjimp.ts and use somewhere in code read from buffer `await Jimp.read(Buffer.from( `
3. use Vite as bundler 
with following configs
...

```
import path from 'path'
import {builtinModules} from 'module'
import {VitePluginNode} from 'vite-plugin-node';
import checker from 'vite-plugin-checker';
// Configure Vitest (https://vitest.dev/config/)
import {defineConfig, ViteDevServer, Connect, loadEnv} from 'vite'

... // relevant part of the vite config
ssr:{  noExternal: true },
plugins: [
  
      checker({
        typescript: true,
      }),
      ...VitePluginNode({
        adapter: 'express',
        appPath: AppMainPath,
        exportName: 'viteNodeApp',
        tsCompiler: 'esbuild',
        swcOptions: {}
      }),
    ],

    test: {
      /* for example, use global to avoid globals imports (describe, test, expect): */
      // globals: true,
    },
    resolve: {
      alias: {
        '@': path.resolve(__dirname, 'src'),
     
      },
    },
    build:
    {
      target: 'esnext',//or module
      ssr: AppMainPath,

      minify: true,// for-dev false
      manifest: true,
      commonjsOptions: {
        transformMixedEsModules: true,
      },
      rollupOptions: {
        input: AppMainPath, // in case we have problem with some deps,
        // but rather remove them to make clean build!
        external: builtinModules
      },
      outDir: './dist'
    }
  }
```
5. create build using vite - `vite build --ssr`
6. run vite preview - of that build - there are no externals so jimp gets compiled in bundle  
  or run the file build directly using node like: node dist/server.mjs
5. you get it... 

**Screenshots**

using gimp in backend app, hosted via vite with typescript option, es module compilation fails although using jimp/es

- Jimp Version: 0.22.7
- Operating System: linux
- Node version:18.3

## Failure Logs
ReferenceError: __dirname is not defined in ES module scope
    at file:///..app/dist/server.mjs:474:21033
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)

### sugested fix  - in your source - quite straightforward
https://stackabuse.com/bytes/fix-dirname-is-not-defined-in-es-module-scope-in-javascript-node/

Javier-Rotelli:
Fixes #1235

# What's Changing and Why
Removed usage of global variable `__dirname` to avoid errors when bundling with rollup

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`

### DIFF
--- a/packages/plugin-print/src/index.js
+++ b/packages/plugin-print/src/index.js
@@ -1,7 +1,11 @@
 import Path from "path";
+import Url from 'url';
 import bMFont from "load-bmfont";
 import { isNodePattern, throwError } from "@jimp/utils";
 import { measureText, measureTextHeight, splitLines } from "./measure-text";
+
+const __filename = url.fileURLToPath(import.meta.url);
+const __dirname = Path.dirname(__filename);
 
 function xOffsetBasedOnAlignment(constants, font, line, maxWidth, alignment) {
   if (alignment === constants.HORIZONTAL_ALIGN_LEFT) {


packages/plugin-print/src/index.js
```diff
@@ -1,8 +1,12 @@
 import Path from "path";
+import Url from 'url';
 import bMFont from "load-bmfont";
 import { isNodePattern, throwError } from "@jimp/utils";
 import { measureText, measureTextHeight, splitLines } from "./measure-text";
 
+const __filename = url.fileURLToPath(import.meta.url);
```
hipstersmoothie, line 7: Does all compiled output have the correct code here?

ex:

- cjs
- esm
- browser

Once verified will merge

hipstersmoothie (changes requested):

hipstersmoothie:
We would have to make sure it works for all envs, not just modules

hipstersmoothie:
Fix in https://github.com/jimp-dev/jimp/pull/1310

Issue #980, opened 2021-01-12
The "chunk" argument must be of type string or an instance of Buffer or Uint8Array. Received an instance of Promise


proginter:
# Expected Behavior
I need it will not kill the program because of an error I dont know why is happening

# Current Behavior
I get sometimes error, I dont know when, but it not on my code.

events.js:292
      throw er; // Unhandled 'error' event
      ^
TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be of type string or an instance of Buffer or Uint8Array. Received an instance of Promise
    at WriteStream.Writable.write (internal/streams/writable.js:285:13)
    at WriteStream.<anonymous> (/home/XXX/domains/XXX.com/public_html/server/node_modules/@jimp/core/dist/index.js:589:18)
    at WriteStream.emit (events.js:315:20)
    at internal/fs/streams.js:380:12
    at FSReqCallback.oncomplete (fs.js:171:23)
Emitted 'error' event on process instance at:
    at emitUnhandledRejectionOrErr (internal/event_target.js:545:11)
    at MessagePort.[nodejs.internal.kHybridDispatch] (internal/event_target.js:358:9)
    at MessagePort.exports.emitMessage (internal/per_context/messageport.js:18:26) {
  code: 'ERR_INVALID_ARG_TYPE'
}

# Failure Information (for bugs)
An error, it happening 1 - 2 times in a day I dont know why, but I using your library to resize and compress images.

## Steps to Reproduce

**Screenshots**

## Context

- Jimp Version: 0.13.0
- Operating System: Centos 7 64bit
- Node version: 14.15.4

## Failure Logs

Thank you!

marinoandrea:
I am getting the same error on version `0.16.1` and node `15.8.0` while performing `Jimp.writeAsync`. 
The error must be inside some callback since try-catching does not help.

agincel:
I'm encountering this same issue.

WannaBeChay:
Same issue.

TunjiDev:
Same issue, how do we resolve this pls?

WannaBeChay:
U are talking about, my problem on stackoverflow ?

Em ter., 3 de ago. de 2021 às 11:27, Igbatayo Adetunji <
***@***.***> escreveu:

flixcheck:
[...]

Issue #1056, opened 2022-01-04
getBase64Async returning stringified Promise


nolmsted:
When using the getBase64Async function, a string like the following is returned 
`data:image/gif;base64,[object Promise]`

Looking at the source at this link ->
https://github.com/oliver-moran/jimp/blob/53ff9d1266207f7f674233f465ec358274510511/packages/core/src/index.js#L595

I'm guessing that the getBase64Async function is still calling the synchronous getBuffer()  rather than the async version. I was able to work around the issue with the following 

`image.getBufferAsync(image.getMIME()).then(data => {
       src = 'data:' + image.getMIME() + ';base64,' + data.toString('base64');`

rdobda-gaia:
After upgrading to node 16, getBufferAsync is timing out on large files. I tried switching to getBase64Async then building a buffer from that result and it mostly worked until i hit this issue, which oddly only happened in some cases and not in others. If the solution is to go back to getBufferAsync then it's an infinite loop of problems.

hipstersmoothie:
@nolmsted would you mind submitting a fix for this?

hipstersmoothie:
:rocket: Issue was released in [`v1.0.2`](https://github.com/jimp-dev/jimp/releases/tag/v1.0.2) :rocket:

PR #1239, opened 2023-05-29
Encoder for GIF provides a Promise to getBuffer when a string, Buffer, or Uint8Array is expected

stevezac-osu:
Fixing Issue 980: The 'chunk' argument must be of type string or an Instance of Buffer or Uint8Array. Received an instance of Promise.

# What's Changing and Why
On jimp/packages/core/src/utils/image-bitmap.js getBuffer(mime, cb) function, the method is intended to provide a string or a buffer to a callback function so that the callback can consume such buffer (typically with the Write function).

However, occasionally the following error will occur ( see also #980 ):

```
internal/streams/writable.js:285
      throw new ERR_INVALID_ARG_TYPE(
      ^

TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be of type string or an instance of Buffer or Uint8Array. Received an instance of Promise
    at new NodeError (internal/errors.js:322:7)
    at WriteStream.Writable.write (internal/streams/writable.js:285:13)
    at WriteStream.<anonymous> (/home/zachs1590/cs464-jimp-testing/node_modules/@jimp/core/dist/index.js:462:16)
    at WriteStream.emit (events.js:400:28)
    at internal/fs/streams.js:340:12
    at FSReqCallback.oncomplete (fs.js:180:23) {
  code: 'ERR_INVALID_ARG_TYPE'
```

I traced the issue back to the fact that one of the encoders, the GIF encoder, consistently returns a promise rather than a string or a buffer response.  See jimp/packages/type-gif/src/index.js line 34.

Rather than solve this issue at the GIF encoder (if that could be solved at all), it seemed like a better idea to make the get buffer function more robust by permitting encoders to return a promise as a valid data type and handle such a response gracefully.

Accordingly, on jimp/packages/core/src/utils/image-bitmap.js getBuffer(mime, cb), I have added

```js
  if (this.constructor.encoders[mime]) {
    const buffer = this.constructor.encoders[mime](this);
    //Typically, buffers return a string or map.  However, the gif library "gifwrap" seemingly returns promises.
    if(buffer instanceof Promise){
      //trigger the callback when the promise has been resolved
      buffer.then((buff) => {
        cb.call(this, null, buff);
      })
    } else {
      cb.call(this, null, buffer);
    }
```
n/a

### DIFF
--- a/packages/core/src/utils/image-bitmap.js
+++ b/packages/core/src/utils/image-bitmap.js
@@ -225,7 +225,15 @@ export function getBuffer(mime, cb) {
 
   if (this.constructor.encoders[mime]) {
     const buffer = this.constructor.encoders[mime](this);
-    cb.call(this, null, buffer);
+    //Typically, buffers return a string or map.  However, the gif library "gifwrap" seemingly returns promises.
+    if(buffer instanceof Promise){
+      //trigger the callback when the promise has been resolved
+      buffer.then((buff) => {
+        cb.call(this, null, buff);
+      })
+    } else {
+      cb.call(this, null, buffer);
+    }
   } else {
     return throwError.call(this, "Unsupported MIME type: " + mime, cb);
   }
--- a/packages/core/src/utils/image-bitmap.js
+++ b/packages/core/src/utils/image-bitmap.js
@@ -225,12 +225,12 @@ export function getBuffer(mime, cb) {
 
   if (this.constructor.encoders[mime]) {
     const buffer = this.constructor.encoders[mime](this);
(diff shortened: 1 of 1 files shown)


stevezac-osu:
Please note -- I also think this solves #1056

jasonz1987:
please merge this pr.

jbwebtech:
@stevezac-osu  thank you for finding (and fixing!) the issue many of us are experiencing.  I implemented your change locally, and it worked great!  👍🎉

PR #1290, opened 2024-05-26
Replace Lenna test image.

scosman:
# What's Changing and Why

Replace Lenna test image.

See: https://en.wikipedia.org/wiki/Lenna#Criticism for context

Discussion/prior okay here: https://github.com/jimp-dev/jimp/issues/711

New image is royalty free from Kodak: https://github.com/lemire/kodakimagecollection

I'm no expert in this project, but seems pretty safe

## Tasks

- [ x ] Add tests -- no additions, but updated tests to use new image.
- [ x ] Update Documentation
- [ na ] Update `jimp.d.ts`
- [ na ] Add [SemVer](https://semver.org/) Label - not required as doesn't change functionality, just docs/tests


hipstersmoothie:
Made sure to remove this in the v1. Thanks for your patience!

PR #1307, opened 2024-09-01
Add quantize plugin

### LINKED ISSUE #871: Reduce the number of colors in an image
I would like to have a method on Jimp to reduce the number of colors in an image. Similar to what happens when we convert an image from RGB to Indexed color / pallet.

It would be nice to have something like:
```
image.reduceColors(numOfColors)
```

The `RgbQuant.js` library does this, but it uses the canvas to accomplish the same effect.

hipstersmoothie:
closes #871


hipstersmoothie:
:rocket: PR was released in [`v1.1.0`](https://github.com/jimp-dev/jimp/releases/tag/v1.1.0) :rocket:

PR #1310, opened 2024-09-02
Fix using dirname in print plugin

hipstersmoothie:
This is the last place in non-test code that __dirname is used. After this esm usage of all apis should be good.


hipstersmoothie:
:rocket: PR was released in [`v1.1.2`](https://github.com/jimp-dev/jimp/releases/tag/v1.1.2) :rocket:

PR #1330, opened 2024-09-07
add JimpInstance type

### LINKED ISSUE #1328: Best practice for "generic" Jimp instance type missing from docs?
**Is your feature request related to a problem? Please describe.**
I'm working on a solution that will load an image from url or a buffer, transform it and output it in base64. This means it will be passed around through a couple functions. I've tried to figure out a type for the Jimp object being passed as parameter, but nothing I tried works (see examples below). Resources from the internet show examples with older versions of Jimp (with default import). 

AFAIK it was possible with older versions just by doing this:
```typescript
import Jimp from "jimp"

function transform(jimpImage: Jimp) {
//...
}
```

 Is it not possible anymore and I should treat url and buffer inputs as completely separate pipelines?

**Describe the solution you'd like**
Hopefully it's just a case of docs not being clear?

**Additional context**

[Sandbox](https://codesandbox.io/p/sandbox/pc26m9)
```typescript
import { Jimp } from "jimp";

// 'Jimp' refers to a value, but is being used as a type here.
async function transform(jimpImage: Jimp) {
  return await jimpImage.getBase64("image/png");
}

// Property 'getBase64' does not exist
async function transform1(jimpImage: typeof Jimp) {
  return await jimpImage.getBase64("image/png");
}

//Using just one return type will not accept the output of the other function
//Results in:
// This expression is not callable. Each member of the union type [...] has signatures, but none of those signatures are compatible with each other.
async function transform2(
  jimpImage:
    | Awaited<ReturnType<typeof Jimp.read>>
    | Awaited<ReturnType<typeof Jimp.fromBuffer>>
) {
  return await jimpImage.getBase64("image/png");
}
```

hipstersmoothie:
closes #1328

### DIFF
--- a/packages/jimp/src/index.ts
+++ b/packages/jimp/src/index.ts
@@ -128,6 +128,8 @@ export const Jimp = createJimp({
   plugins: defaultPlugins,
 });
 
+export type JimpInstance = InstanceType<typeof Jimp>;
+
 export type {
   ColorAction,
   HueAction,


hipstersmoothie:
:rocket: PR was released in [`v1.4.0`](https://github.com/jimp-dev/jimp/releases/tag/v1.4.0) :rocket:

PR #1336, opened 2024-09-07
Add support for image decoder options

### LINKED ISSUE #939: Unable to configure decoder options
https://github.com/oliver-moran/jimp/blob/909f06eac36ef5ba36ebb1e0853fd006c4da3560/packages/core/src/utils/image-bitmap.js#L165

This library doesn't allow for configuring some decoders that take extra parameters. The line above shows this explicitly where only one argument is given instead of at least another options object. Decoders such as the `jpeg-js` one offer extra configurations in the form of another configuration that can be useful for a given context (see here google/eleventy-high-performance-blog#10).

My proposal fix would be to change the constructor function to allow an `decoderOptions` key that map mimetypes to the respective decoder options. For example:
```js
const decoderOptions = {
  "image/jpeg": {
    maxMemoryUsageInMB: 1024,
  }
};
Jimp.decode({ path: "filepath.jpg", decoderOptions, });
Jimp.decode("filepath.jpg", { decoderOptions }); // alternative
```

hipstersmoothie:
closes #939

## Release Notes

Can now have options for the underlying image codecs

CleanShot 2024-09-07 at 15 26 41

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -98,13 +98,20 @@ type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends (
 type Constructor<T> = new (...args: any[]) => T;
 
 type JimpFormat<
-  M extends string = string,
-  O extends Record<string, any> | undefined = undefined,
-  T extends Format<M, O> = Format<M, O>,
+  MimeType extends string = string,
+  EncodeOptions extends Record<string, any> | undefined = undefined,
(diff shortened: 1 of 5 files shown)


hipstersmoothie:
:rocket: PR was released in [`v1.6.0`](https://github.com/jimp-dev/jimp/releases/tag/v1.6.0) :rocket:

PR #1367, opened 2024-12-26
docs: add share image and favicon for the docs site

crutchcorn:
# What's Changing and Why

This PR adds a share image for social media sites and a favicon that's not the generic starlight favicon. Just some small polishing touches on the docs site

Nothing

## Tasks

- [ ] Add tests
- [ ] Update Documentation
- [ ] Update `jimp.d.ts`


packages/docs/src/overwrite-components/head.astro
```diff
@@ -0,0 +1,11 @@
+---
+import type { Props } from "@astrojs/starlight/props";
+import Default from "@astrojs/starlight/components/Head.astro";
+---
+
+<Default {...Astro.props}>
+	<slot />
+</Default>
+<meta name="twitter:image" content="https://jimp-dev.github.io/jimp/share-banner.png">
```
crutchcorn, line 9: These have to be a static URL, otherwise some sites don't know how to resolve this quite right. If we add a custom domain in the future, we'll want to update this

PR #1383, opened 2025-05-10
fix: load bitmap font fail in browser

xdoer:
Jimp supports loading fonts from URLs, but on the browser, I found that it performs abnormally under non localhost domains. 

Then I found that there was a problem with the following line of code,
https://github.com/jimp-dev/jimp/blob/b6b0e418a5f1259211a133b20cddb4f4e5c25679/plugins/plugin-print/src/load-bitmap-font.ts#L161

and it will eventually construct a similar one `https:/raw.githubusercontent.com/jimp-dev/jimp/main/plugins/plugin-print/fonts/open-sans/open-sans-16-black/open-sans-16-black.png`  

The link clearly shows that there is a missing '/' after 'https'.

### DIFF
--- a/plugins/plugin-print/src/load-bitmap-font.ts
+++ b/plugins/plugin-print/src/load-bitmap-font.ts
@@ -157,9 +157,15 @@ export async function processBitmapFont(file: string, font: LoadedFont) {
     chars,
     kernings,
     pages: await Promise.all(
-      font.pages.map(async (page) =>
-        CharacterJimp.read(path.join(path.dirname(file), page))
-      )
+      font.pages.map(async (page) => {
+        let fileUrl = ''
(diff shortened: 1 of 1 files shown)


xdoer:
I found it is interesting, I used the following code to test it separately in Node environment, browser environment under localhost URL, and https URL. 

```javascript
fetch("https:/raw.githubusercontent.com/jimp-dev/jimp/main/plugins/plugin-print/fonts/open-sans/open-sans-16-black/open-sans-16-black.png")
```

It can take effect in both Node environment and browser localhost address. 
But it does not take effect in  HTTPS address.

BigYellowHammer:
@hipstersmoothie Any chance to take a look on that please?

